PR #7, opened 2014-04-12
Security model: add sandboxing behaviour and mention fullscreen as MAY

mounirlamouri:


index.html
```diff
@@ -619,6 +634,43 @@
...
+          The <dfn>sandboxed orientation lock browsing context flag</dfn>
+        </dt>
+        <dd>
+          This flag disables the ability to lock the screen orientation.
+        </dd>
+      </dl>
+      <p>
+        The algorithm to <a href=
+        'http://www.w3.org/TR/html5/browsers.html#parse-a-sandboxing-directive'>
+        parse a sandboxing directive</a> is also extended in step 3 such as:
```
marcoscaceres: - such that

index.html
```diff
@@ -619,6 +634,43 @@
         Documents</a>.
       </p>
     </section>
+    <section>
+      <h2>
+        Extension to the HTML sandboxing model
+      </h2>
+      <p>
```
marcoscaceres, line 641: Put a note here to let people know that this is supposed to be in the HTML spec.

PR #9, opened 2014-04-12
Handle the case where locking the screen orientation is not supported.

mounirlamouri:


index.html
```diff
@@ -304,8 +304,9 @@
           able to lock the screen orientation.
         </p>
         <p>
-          The screen orientation MUST be lockable to all of the following
-          values:
+          The <a>user agent</a> MAY not support locking the screen orientation
```
marcoscaceres: "MAY not" is ugly. Please change it to: "It is OPTIONAL for a user agent to support locking the screen orientation." Also, optional conditional (OPTIONAL... MUST...) are a bit eeky. We can probably fix this later tho.

index.html
```diff
@@ -347,6 +348,9 @@
           <var>orientation</var> are given by the following algorithm:
         </p>
         <ol>
+          <li>If the locking the screen orientation is not supported, return a
```
marcoscaceres: Use active voice instead of passive (i.e., if the user agent does not support screen orientation). The way the sentence is currently written is kinda wrong: "the locking" is grammatically incorrect.

index.html
```diff
@@ -347,6 +348,9 @@
...
+          Promise rejected with a <code>DOMException</code> whose name is <code>
+          NotSupportedError</code> and abort these step.</li>
```
marcoscaceres: as above, make it active voice. "Reject promise with DOMException"

mounirlamouri:
Marcos, could you please take another look?

marcoscaceres:
lgtm

PR #15, opened 2014-06-06
Began addressing feedback from bug #24697

marcoscaceres:
Started rewriting steps to apply an orientation lock algorithm as per Bugzilla bug [24697](https://www.w3.org/Bugs/Public/show_bug.cgi?id=24697). 

It reworks the way the promise are resolved/rejected and the "pending promise" which is in the asynchronous task.

### DIFF
--- a/index.html
+++ b/index.html
@@ -352,7 +352,11 @@
         </dl>
         <p>
           The steps to <dfn>apply an orientation lock</dfn> using
-          <var>orientation</var> are given by the following algorithm:
+          <var>orientation</var> are given by the following algorithm. The
+          algorithm assumes that for each <a>top-level browsing context</a>
+          there is a <var>pending promise</var>, which is initially set to
+          <var>null</var>, which represents a pending task to lock the screen
+          orientation:
         </p>
         <ol>
           <li>If the <a>user agent</a> does not support locking the screen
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -352,29 +356,25 @@
...
               </li>
               <li>If <var>doc</var> has an empty <a>fullscreen element
-              stack</a>, return a Promise rejected with a
+              stack</a>, reject <var>promise</var> with a
               <code>DOMException</code> whose name is
-              <code>SecurityError</code> and abort these steps.
+              <code>SecurityError</code>.
               </li>
-              <li>When <var>doc</var> <a>fully exit fullscreen</a>, then
+              <li>When <var>doc</var> <a>fully exits fullscreen</a>, then
```
mounirlamouri: You will need to update the `<dfn>` to be "fully exits fullscreen".

index.html
```diff
@@ -412,9 +412,20 @@
           <li>Otherwise, append <var>orientation</var> to
           <var>orientations</var>.
           </li>
+          <li>If <var>pending promise</var> is not <code>null</code>:
+            <ol>
+              <li>Reject <var>pending-promise</var> with a
+              <code>DOMException</code> whose name is <code>AbortError</code>.
```
mounirlamouri: I guess you can have the entire thing in one line:

``` html
<li>If <var>pending promise</var> is not <code>null</code>, reject <var>pending-promise</var> with
<code>DOMException</code> whose name is <code>AbortError</code>.</li>
```

This said, wouldn't that be simpler to have that on top? Before the fullscreen check for example. Then, the fullscreen check rejection could abort the steps. Otherwise, it's a bit painful to keep a promise around that might or might not be rejected. WDYT?

index.html
```diff
@@ -412,9 +412,20 @@
...
+              </li>
+            </ol>
+          </li>
+          <li>Set <var>pending promise</var> to be <var>promise</var>.
+          </li>
           <li>
             <a>Queue a task</a> to perform the following substeps:
             <ol>
+              <li>If <var>promise</var> has been rejected, abort these steps.
```
mounirlamouri: I don't think you would need that if you make the pending-promise check happen before the fullscreen check. Note that you were missing a `return <var>promise</var>`.

mounirlamouri:
Thanks for this :) And sorry for not looking into it earlier. I wanted to tackle some low hanging fruit before tackling the more complex ones ;)
Could you update your PR to apply on top of tree and have my comments taken into account. I will have another look after.

index.html
marcoscaceres: fixed.

index.html
marcoscaceres: Agreed. Will fix.

index.html
marcoscaceres: Returning the promise happened at the end of the algo. However, I've changed this as you suggested.

marcoscaceres:
I made the changes. You were correct about the dangling promise - sorry about that. However, there is still a big issue with the following text:

> Otherwise, depending on platform conventions, the user agent MAY change how the viewport is drawn in order to make it match another screen orientation type or angle. However, it MUST be part of orientations. 

It's obviously not good to leave this like this - and the text above is really confusing for me. 

I think how we can fix the whole thing is to make sure "Lock the orientation" returns either true or false. That way,  we can properly specify the whole screen orientation change as part of the "task" that has been queued. This includes:
1. Rejecting if orientation change failed + clean up the _pending promise_ right there and then.  
2. Resolving with the `OrientationInformation` (even if nothing changed) and clean up pending promise. 
3. Queue a task to fire the event. This needs a different task queue, btw - because we need to be able to trash any pending tasks that are wanting to change the screen orientation independently of orientation change events.

index.html
```diff
@@ -360,34 +364,41 @@
           <code>DOMException</code> whose name is
           <code>NotSupportedError</code> and abort these steps.
           </li>
+          <li>If <var>pending promise</var> is not <code>null</code>:
+            <ol>
+              <li>Cancel all tasks in the <a>screen orientation task
+              source</a>.
```
mounirlamouri, line 370: Why? I'm not sure what you are trying to do here.

index.html
```diff
@@ -360,34 +364,41 @@
...
             <ol>
               <li>Let <var>doc</var> be the <a>top-level browsing context</a>'s
-              document.
+              <code>document</code>.
               </li>
               <li>If <var>doc</var> has an empty <a>fullscreen element
               stack</a>, return a Promise rejected with a
               <code>DOMException</code> whose name is
-              <code>SecurityError</code> and abort these steps.
+              <code>SecurityError</code>.
```
mounirlamouri: Why not breaking the steps anymore?

index.html
marcoscaceres, line 370: Presumably we are about to perform an orientation change, so we need to clobber all pending tasks that are wanting to change the orientation (there should only ever be one task in waiting in the event loop).

index.html
marcoscaceres: Added it back in.

index.html
mounirlamouri, line 370: Ok. What bothers me is that you can't really prevent another lock to apply because there is a system layer that might not be controllable but I guess this is not the intent here.

mounirlamouri:
LGTM. Thanks! :)

PR #16, opened 2014-06-23
Define what the term 'screen' means in the context of the specification.

mounirlamouri:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=25311


mounirlamouri:
@marcoscaceres can you have a look at this?

marcoscaceres:
LGTM

PR #18, opened 2014-06-23
Remove ability to lock to angles and add ability to lock to 'natural' orientation.

mounirlamouri:
Solves the following bug: https://www.w3.org/Bugs/Public/show_bug.cgi?id=25329

This push has two commits:
"Remove the ability to lock to angles." that removes the support for locking to 0, 90, 180 and 270.
"Add 'natural' orientation lock type." that adds support for locking to 'natural' which is equivalent to locking to angle 0.

Please ignore the following commit:
"Define what the term 'screen' means in the context of the specification."

### DIFF
--- a/index.html
+++ b/index.html
@@ -470,7 +470,7 @@
               <i>orientations</i>.
               </li>
               <li>The <a>user agent</a> MUST change how the viewport is drawn
-              such as the <a>current screen orientation type</a> will be equal
+              so that the <a>current screen orientation type</a> will be equal
               to <i>orientation</i>.
               </li>
               <li>After the change has happened, the <a>user agent</a> MUST
--- a/index.html
+++ b/index.html
@@ -329,6 +329,9 @@
             any
           </dt>
           <dt>
+            natural
+          </dt>
+          <dt>
             landscape
           </dt>
           <dt>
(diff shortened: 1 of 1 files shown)


mounirlamouri:
@marcoscaceres can you take a look at this? :)

index.html
```diff
@@ -464,9 +475,8 @@
               <i>orientations</i>.
               </li>
               <li>The <a>user agent</a> MUST change how the viewport is drawn
-              such as the <a>current screen orientation type</a> or the
-              <a>current screen orientation angle</a> will be equal to
-              <i>orientation</i>.
+              such as the <a>current screen orientation type</a> will be equal
```
marcoscaceres: - such that

marcoscaceres: Or "so that"

index.html
```diff
@@ -476,12 +486,10 @@
...
           </li>
-          <li>If neither the <a>current screen orientation type</a> nor the <a>
-            current screen orientation angle</a> values are part of
-            <i>orientations</i>, the <a>user agent</a> MUST change how the
+          <li>If the <a>current screen orientation type</a> is not part of <i>
+            orientations</i>, the <a>user agent</a> MUST change how the
             viewport is drawn such as the <a>current screen orientation
-            type</a> or the <a>current screen orientation angle</a> will be
-            equal to one of <i>orientations</i>' values.
+            type</a> will be equal to one of <i>orientations</i>' values.
```
marcoscaceres, line 487: Is that closing `</a>` there a stray tag?

index.html
mounirlamouri: Done

index.html
mounirlamouri, line 487: No. It's closing the <a> opened for "current screen orientation type".

mounirlamouri:
I assume that it is fine to merge. Please open another issue if you had any other comments.

PR #19, opened 2014-06-23
Increase clarity when lockOrientation() locks to current orientation.

mounirlamouri:
Solves issue 12.

### DIFF
--- a/index.html
+++ b/index.html
@@ -425,23 +425,18 @@
           <var>orientations</var>.
           </li>
           <li>
-            <a>Queue a task</a> to perform the following substeps:
-            <ol>
-              <li>
-                <a>Lock the orientation</a> to <var>orientations</var>.
-              </li>
-              <li>Let <var>result</var> be a new
-              <code>OrientationInformation</code>.
-              </li>
-              <li>If locking the orientation does not result in a change of
-              orientation, resolve <var>promise</var> with <var>result</var>
-              and abort these steps.
-              </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -425,23 +425,18 @@
...
+          <li>If <var>orientations</var> contains the <a>current screen
+          orientation type</a>, resolve <var>promise</var> with a new
+          <code>OrientationInformation</code>.
           </li>
-          <li>Otherwise, the next time the orientation changes, before firing
-          the <code>orientationchange</code> event, resolve <var>promise</var>
-          with <var>result</var>.
+          <li>Otherwise, the next time the orientation changes, if the new <a>
+            current screen orientation type</a> is included in
+            <var>orientations</var>, before firing the
```
marcoscaceres, line 437: If you are certain that you can resolve promises before events are fired, then all this is fine. I don't know how this works internally, tbh: I assumed that putting the "task" in the particular task queue would imply the order in which the promise is resolved and the event is fired. There is a explicit task queue for this, right?

mounirlamouri, line 437: Yes, there is one, see 4.4: "Unless specified otherwise, when queuing a task, the screen orientation task source is used.".

Maybe the task queue would handle the order by itself but making sure it's specified doesn't hurt ;)

marcoscaceres, line 437: Excellent. Merge this bad boy! :feelsgood:

Issue #11, opened 2014-04-12
OrientationInformation passed to Promise as a result of lockOrientation is broken


mounirlamouri:
It should be frozen, somehow.

marcoscaceres:
Let's make it a dictionary as you suggested.

PR #20, opened 2014-07-03
Do not resolve the locking promise with OrientationInformation.

mounirlamouri:
More I think about it, less I can see a real UC for that. I think it's probably better to simply better to assume that developers will read `screen.orientation` when the lock is resolved.

Should solve #11.


marcoscaceres:
LGTM

PR #21, opened 2014-07-05
Rewrite to use ScreenOrientation interface (closes #13)

### LINKED ISSUE #13: OrientationInformation should be a dictionary
It is not a constructible class, and has no methods or `this`-dependent constructs, and so should just be a normal JavaScript object (i.e. "dictionary" in WebIDL terms).

marcoscaceres:
- added ScreenOrientation interface
- Rewrote examples
- Removed redundant "dependencies"
- Merged dependencies and terminology section
- Made Promise linkable in dependencies
- Moved ScreenOrientation and Extension to navigator object to the top
  of the document
- Removed "concepts" section (left all inner content, of course)
- rewrote how to get attributes
- removed redundancy when invoking lock() and unlock()
- Moved enums to their  own sections (it was annoying trying to find
  them. Now the show up in the TOC!)
- Introduced "security condition" concept. This fixes the weird
  dependency on the FULLSCREEN API. That's now moved to the "Interaction
  with the FullScreen API" section
-  Added Interaction with FullScreen API section. Defined what to do
  when full screen fully exits there.
- moved conformance section to bottom of document

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,6 +31,11 @@
               companyURL: "http://google.com/",
               note: "previously Mozilla",
             },
+            {
+              name: "Marcos Cáceres",
+              company: "Mozilla",
+              companyURL: "http://mozilla.com/",
+            }
         ],
 
         wg:           "Web Applications Working Group",
@@ -44,8 +49,8 @@
     <section id='abstract'>
       The <cite>Screen Orientation API</cite> provides the ability to read the
       screen orientation type and angle, to be informed when the screen
-      orientation state changes and be able to lock the screen orientation to a
-      specific state.
+      orientation state changes, and be able to lock the screen orientation to
+      a specific state.
     </section>
     <section id='sotd'>
       This document is still in a work in progress state. You can have a look
@@ -66,29 +71,24 @@
         changes:
       </p>
       <pre class='example highlight'>
-        &lt;!DOCTYPE html&gt;
-        &lt;html&gt;
-          &lt;head&gt;
-            &lt;title&gt;Screen Orientation API Example 1&lt;/title&gt;
-            &lt;script&gt;
-              function show() {
-                alert("Screen orientation state is now " + screen.orientation);
-              }
+&lt;script&gt;
+  var show = function() {
+     console.log("Orientation is " + screen.orientation.type);
+  }
+  
+  screen.orientation.addEventListener("change", show);
+  window.onload = show;
+&lt;/script&gt;
 
-              window.addEventListener("orientationchange", show);
-              window.addEventListener("load", show);
-            &lt;/script&gt;
-          &lt;/head&gt;
-          &lt;body&gt;
-            &lt;input type='button' value='Unlock'
-              onclick='screen.unlockOrientation();'&gt;
-            &lt;input type='button' value='Lock to portrait'
-              onclick="screen.lockOrientation('portrait');"&gt;
-            &lt;input type='button' value='Lock to landscape'
-              onclick="screen.lockOrientation('landscape');"&gt;
-          &lt;/body&gt;
-        &lt;/html&gt;
-
+&lt;button onclick='screen.orientation.unlock()'&gt;
+  Unlock
+&lt;/button&gt;
+&lt;button onclick="screen.orientaiton.lock('portrait')"&gt;
+  Lock to portrait
+&lt;/button&gt;
+&lt;button onclick="screen.orientation.lock('landscape')"&gt;
+  Lock to landscape
+&lt;/button&gt;
 </pre>
     </section>
     <section id='conformance'>
@@ -108,38 +108,28 @@
         Dependencies
       </h2>
       <p>
-        The following concepts, terms and elements are defined in [[!HTML5]]:
+        The following concepts and interfaces are defined in [[!HTML5]]:
       </p>
       <ul>
         <li>
           <dfn><a href=
-          "http://dev.w3.org/html5/spec/webappapis.html#queue-a-task">queue a
+          "http://www.w3.org/TR/html5/webappapis.html#queue-a-task">queue a
           task</a></dfn>
         </li>
         <li>
           <dfn><a href=
-          "http://dev.w3.org/html5/spec/webappapis.html#fire-a-simple-event">fire
+          "http://www.w3.org/TR/html5/webappapis.html#fire-a-simple-event">fire
           a simple event</a></dfn>
         </li>
         <li>
           <dfn><a href=
-          "http://dev.w3.org/html5/spec/webappapis.html#event-handlers">event
+          "http://www.w3.org/TR/html5/browsers.html#event-handlers">event
           handler</a></dfn>
         </li>
         <li>
           <dfn><a href=
-          "http://dev.w3.org/html5/spec/webappapis.html#event-handler-event-type">
-          event handler event type</a></dfn>
-        </li>
-        <li>
-          <dfn><a href=
-          '%20http://www.w3.org/TR/html5/webappapis.html#event-handler-idl-attributes'>
-          event handler IDL attribute</a></dfn>
-        </li>
-        <li>
-          <dfn><a href=
-          'http://www.w3.org/TR/html5/webappapis.html#event-handler-content-attributes'>
-          event handler content attribute</a></dfn>
+          "http://www.w3.org/TR/html5/browsers.html#event-handler-event-type">event
+          handler event type</a></dfn>
         </li>
         <li>
           <code><dfn><a href=
@@ -147,29 +137,16 @@
         </li>
         <li>
           <code><dfn><a href=
-          '%20http://www.w3.org/TR/html5/dom.html#document'>Document</a></dfn></code>
-        </li>
-        <li>
-          <code><dfn><a href=
-          'http://www.w3.org/TR/html5/sections.html#the-body-element'>body</a></dfn></code>
-        </li>
-        <li>
-          <code><dfn><a href=
-          '%20http://www.w3.org/TR/html5/obsolete.html#frameset'>frameset</a></dfn></code>
+          'http://www.w3.org/TR/html5/dom.html#document'>Document</a></dfn></code>
         </li>
         <li>
           <dfn><a href=
-          '%20http://www.w3.org/TR/html5/browsers.html#concept-document-window'>
-          Window object's Documents</a></dfn>
-        </li>
-        <li>
-          <dfn><a href=
-          "http://dev.w3.org/html5/spec/browsers.html#browsing-context">browsing
+          "http://www.w3.org/TR/html5/browsers.html#browsing-context">browsing
           context</a></dfn>
         </li>
         <li>
           <dfn><a href=
-          "http://www.w3.org/html/wg/drafts/html/master/browsers.html#top-level-browsing-context">
+          "http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context">
           top-level browsing context</a></dfn>
         </li>
         <li>browsing context's <dfn><a href=
@@ -190,7 +167,7 @@
       <p>
         <a href=
         'http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects'>
-        Promise objects</a> are defined in [[!ECMASCRIPT]].
+        <dfn>Promise</dfn> objects</a> are defined in [[!ECMASCRIPT]].
       </p>
       <p>
         <dfn><a href=
@@ -202,23 +179,33 @@
     </section>
     <section>
       <h2>
-        Concepts
+        Interface definitions
       </h2>
       <section>
         <h2>
-          Terminology
-        </h2>
-        <p>
-          Unless stated otherwise, the term <dfn>screen</dfn> is equivalent to
-          the screen of the output device associated to the <a>Window</a>, as
-          per [[!CSSOM-VIEW]].
-        </p>
+          Extensions to the <a>Screen</a> interface
+        </h2>The CSSOM View specification defines a <code>Screen</code>
+        interface [[!CSSOM-VIEW]], which this specification extends:
+        <dl class='idl' title='partial interface Screen'>
+          <dt>
+            readonly attribute ScreenOrientation orientation
+          </dt>
+        </dl>
       </section>
       <section>
         <h2>
-          Reading the screen orientation
+          <code>ScreenOrientation</code> interface
         </h2>
-        <dl title="interface OrientationInformation" class="idl">
+        <dl class='idl' title='interface ScreenOrientation : EventTarget'>
+          <dt>
+            Promise&lt;undefined&gt; lock(OrientationLockType orientation)
+          </dt>
+          <dt>
+            void unlock()
+          </dt>
+          <dt>
+            attribute EventHandler onchange
+          </dt>
           <dt>
             readonly attribute OrientationType type
           </dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -66,49 +66,31 @@
...
-        &lt;!DOCTYPE html&gt;
-        &lt;html&gt;
-          &lt;head&gt;
-            &lt;title&gt;Screen Orientation API Example 1&lt;/title&gt;
-            &lt;script&gt;
-              function show() {
-                alert("Screen orientation state is now " + screen.orientation);
-              }
+&lt;script&gt;
+  var show = () =&gt; {
```
mounirlamouri: Maybe we should avoid fancy syntax and simply have:

``` js
function show() {
  ...
}
```

index.html
```diff
@@ -66,49 +66,31 @@
...
-            &lt;input type='button' value='Unlock'
-              onclick='screen.unlockOrientation();'&gt;
-            &lt;input type='button' value='Lock to portrait'
-              onclick="screen.lockOrientation('portrait');"&gt;
-            &lt;input type='button' value='Lock to landscape'
-              onclick="screen.lockOrientation('landscape');"&gt;
-          &lt;/body&gt;
-        &lt;/html&gt;
+&lt;button onclick='screen.orientation.unlock()'&gt;
+  Unlock
```
mounirlamouri, line 84: You forgot `</button>`.

index.html
```diff
@@ -66,49 +66,31 @@
...
 
+&lt;button onclick="screen.orientaiton.lock('portrait')"&gt;
+  Lock to portrait
```
mounirlamouri, line 87: ditto

index.html
```diff
@@ -66,49 +66,31 @@
...
+
+&lt;button onclick="screen.orientation.lock('landscape')"&gt;
+  Lock to landscape
```
mounirlamouri, line 90: ditto

index.html
```diff
@@ -123,25 +105,10 @@
...
-        </li>
-        <li>
-          <dfn><a href=
-          "http://dev.w3.org/html5/spec/webappapis.html#event-handler-event-type">
-          event handler event type</a></dfn>
-        </li>
-        <li>
-          <dfn><a href=
-          '%20http://www.w3.org/TR/html5/webappapis.html#event-handler-idl-attributes'>
+          "www.w3.org/html/wg/drafts/html/master/webappapis.html#event-handler-idl-attributes">
```
mounirlamouri: You forgot `http://...`

index.html
marcoscaceres: reverted.

index.html
marcoscaceres, line 90: button is self closing, but I added the closing tag back in

index.html
marcoscaceres: fixed

index.html
```diff
@@ -199,418 +153,18 @@
...
         fullscreen</a></dfn> are defined in [[!FULLSCREEN]].
       </p>
-    </section>
-    <section>
-      <h2>
-        Concepts
-      </h2>
-      <section>
-        <h2>
-          Terminology
```
mounirlamouri, line 209: I added this section for `screen` on purpose. I didn't want this to be merged with Dependencies because it is not a dependency, it's a terminology we use in the document.

index.html
```diff
@@ -619,58 +173,435 @@
       [[!CSSOM-VIEW]], which this specification extends:
       <dl class='idl' title='partial interface Screen'>
         <dt>
-          readonly attribute OrientationInformation orientation
+          readonly attribute ScreenOrientation orientation
+        </dt>
+      </dl>
+    </section>
+    <section>
```
mounirlamouri: What about having all the WebIDL sections into a `Interface definitions` section?

index.html
```diff
@@ -619,58 +173,435 @@
...
+        return the <a>current screen orientation angle</a>.
+      </p>
+      <p>
+        The <dfn>onchange</dfn> attribute is an event handler IDL attribute.
       </p>
     </section>
     <section>
       <h2>
-        Extensions to the <a>WindowEventHandlers</a> interface
+        enum <code>OrientationType</code>
```
mounirlamouri: &lt;code>OrientationType&lt;/code> enum

index.html
```diff
@@ -619,58 +173,435 @@
...
+          landscape-primary
+        </dt>
+        <dt>
+          landscape-secondary
         </dt>
       </dl>
+    </section>
+    <section>
+      <h2>
+        enum <code>OrientationLockType</code>
```
mounirlamouri: &lt;code>OrientationLockType&lt;/code> enum

marcoscaceres: fixed

index.html
marcoscaceres: fixed

index.html
```diff
@@ -619,58 +173,435 @@
...
+          landscape-primary
+        </dt>
+        <dt>
+          landscape-secondary
+        </dt>
+      </dl>
+    </section>
+    <section>
+      <h2>
+        Reading the screen orientation
```
mounirlamouri: Would you be against having &lt;section>Concepts&lt;/code> to wrap all the concepts? I think it makes the TOC more readable.

marcoscaceres: I'm ok with it. Will add it.

index.html
marcoscaceres: I don't think you need it (it's not great having sections that are just empty wrappers). But what the hell, we did it for Concepts already so I'll add it :dancer:. Done!

index.html
```diff
@@ -227,14 +204,93 @@
...
+          When getting the <code><dfn id=
+          "widl-ScreenOrientation-type">type</dfn></code> attribute, the user
+          agent MUST return the <a>current screen orientation type</a>.
         </p>
         <p>
-          The <code>angle</code> attribute MUST return the <a>current screen
-          orientation angle</a>.
+          When getting the <code><dfn id=
+          "widl-ScreenOrientation-angle">angle</dfn></code> attribute, the use
+          agent MUST return the <a>current screen orientation angle</a>.
```
mounirlamouri: typo: user agent

index.html
```diff
@@ -310,8 +352,10 @@
         </h2>
         <p>
           The <a>user agent</a> MAY require a <a>document</a>'s <a>top-level
-          browsing context</a> to be fullscreen [[!FULLSCREEN]] in order to be
-          able to lock the screen orientation.
+          browsing context</a> meet one or more <dfn>security conditions</dfn>
+          order to be able to lock the screen orientation. For example, a use
+          agent might require the top-level browsing context to be fullscreen
```
mounirlamouri: typo: user agent

index.html
```diff
@@ -310,8 +352,10 @@
...
+          [[FULLSCREEN]] in order to allow an orientation change.
```
mounirlamouri: in order to allow an orientation **lock**.

index.html
```diff
@@ -716,50 +670,87 @@
...
+        <p>
+          As a <a title="security conditions">security condition</a>, a user
+          agent MAY restrict locking the screen orientation exclusively to when
+          the top-level browsing document's full-screen enabled flag has been
+          set [[!FULLSCREEN]] (i.e., when the document is in fullscreen).
+        </p>
+        <p>
+          If the screen is locked to some orientation when in fullscreen, then,
+          when the document fully exits fullscreen, the user agent MUST lock
+          the orientation to the default orientation.
```
mounirlamouri: Sure. But only if the lock is possible because the document was fullscreen.

index.html
```diff
@@ -716,50 +670,87 @@
...
+          game pointing to an objective.
+        </p>
+        <p>
+          Drawing on the screen in order to point to a physical location
+          requires to know the device orientation and the orientation of the
+          screen in the device coordinates. Without the APIs defined in this
+          specification, a developer has to assume that the <a>current screen
+          orientation angle</a> is 0. With the help of the APIs described in
+          this specification, the developer can <a>apply an orientation
+          lock</a> using <code>0</code> to make that assumption a certitude.
```
mounirlamouri: I know you didn't really changed that, but could you:
s/using &lt;code>0&lt;/code>/using &lt;code>natural&lt;/code>/

index.html
```diff
@@ -227,14 +204,93 @@
...
         </p>
         <p>
+          The <code><dfn id=
+          "widl-ScreenOrientation-onchange">onchange</dfn></code> attribute is
+          an <a>event handler IDL attribute</a>.
+        </p>
+      </section>
+      <section>
+        <h2>
+          enum <code>OrientationType</code>
```
mounirlamouri: Hmm, did you revert the change you did before?
&lt;code>Foo&lt;/code> enum
instead of
enum &lt;code>Foo&lt;/code>

index.html
```diff
@@ -227,14 +204,93 @@
...
+            landscape-primary
+          </dt>
+          <dt>
+            landscape-secondary
+          </dt>
+        </dl>
+      </section>
+      <section>
+        <h2>
+          enum <code>OrientationLockType</code>
```
mounirlamouri: ditto

index.html
```diff
@@ -66,49 +66,32 @@
...
-      </p>
-      <p>
-        Implementations that use ECMAScript to expose the APIs defined in this
-        specification MUST implement them in a manner consistent with the
-        ECMAScript Bindings defined in the Web IDL specification [[!WEBIDL]].
-      </p>
-    </section>
     <section>
       <h2>
         Dependencies
```
mounirlamouri, line 108: I'm still not a big fan of that merge. Can we keep the dependencies and the "terms" in different places?

index.html
```diff
@@ -199,26 +154,48 @@
         'http://www.w3.org/TR/fullscreen/#fully-exit-fullscreen'>fully exit
         fullscreen</a></dfn> are defined in [[!FULLSCREEN]].
```
mounirlamouri, line 177: As long as you remove dependencies, we might want to remove FULLSCREEN from the dependencies. It is only mentioned in non normative requirements so lets not mention it outside so it will be a non-normative dependency. WDYT?

index.html
```diff
@@ -199,26 +154,48 @@
...
       </p>
+      <p>
+        Unless stated otherwise, the term <dfn>screen</dfn> is equivalent to
+        the screen of the output device associated to the <a>Window</a>, as per
+        [[!CSSOM-VIEW]].
+      </p>
+      <p>
+        Algorithms defined in this specification assume that for each
+        <a>top-level browsing context</a> there is a <dfn>pending
+        promise</dfn>, which is initially set to <code>null</code>, which is a
```
mounirlamouri: You are doing <dfn>pending-promise</dfn> but you never actually link to it. Is that on purpose?

index.html
```diff
@@ -227,14 +204,93 @@
...
-          The <code>angle</code> attribute MUST return the <a>current screen
-          orientation angle</a>.
+          When getting the <code><dfn id=
+          "widl-ScreenOrientation-angle">angle</dfn></code> attribute, the use
+          agent MUST return the <a>current screen orientation angle</a>.
         </p>
         <p>
+          The <code><dfn id=
+          "widl-ScreenOrientation-onchange">onchange</dfn></code> attribute is
+          an <a>event handler IDL attribute</a>.
```
mounirlamouri: Hmm, I don't think that's enough. It seems that other specs would do something like:

```
<code><dfn id='widl-ScreenOrientation-onchange'>onchange</dfn></code> is an <a>event handler</a> with the corresponding <code>change</code>
<a>event handler event type</a> that MUST be supported as attribute solely by the
<a>ScreenOrientation</a> object.
```

If you do that, you will have to update the dependencies.

index.html
```diff
@@ -575,22 +582,30 @@
...
+          visible per [[!PAGE-VISIBILITY]], abort these steps.
           </li>
-          <li>If a Promise created in <a>apply an orientation lock</a> steps is
-          pending, it MUST be resolved with <code>undefined</code>.
-          </li>
-          <li>The <a>user agent</a> MUST <a>queue a task</a> to <a>fire a
-          simple event</a> named <code>orientationchange</code> at the
-          <a>browsing context</a> <a>active document</a>'s <a>Window</a> object
-          and wait until that task begins to continue to the next step.
+          <li>If <a>pending promise</a> is not <code>null</code>:
```
mounirlamouri: We should update the pending promise _after_ we update the current angle/type.

index.html
```diff
@@ -575,22 +582,30 @@
...
+            <ol>
+              <li>Resolve <a>pending promise</a> with <code>undefined</code>.
```
mounirlamouri: Not _always_. It should only happen if the new orientation matches the lock.

index.html
```diff
@@ -575,22 +582,30 @@
...
+              </li>
+              <li>Update the <a>current screen orientation angle</a>.
+              </li>
+              <li>
+                <a>Fire a simple event</a> named <code>change</code> at each
+                <a href=
+                "#widl-Screen-orientation"><code>screen.orientation</code></a>
+                object.
+              </li>
+            </ol>
```
mounirlamouri, line 627: Could you add a paragraph making it clear that this mean that a non-visible page will get stalled orientation values which will be updated when the page will be visible again - with a change event.

index.html
```diff
@@ -310,8 +352,10 @@
         </h2>
         <p>
           The <a>user agent</a> MAY require a <a>document</a>'s <a>top-level
-          browsing context</a> to be fullscreen [[!FULLSCREEN]] in order to be
-          able to lock the screen orientation.
+          browsing context</a> meet one or more <dfn>security conditions</dfn>
+          order to be able to lock the screen orientation. For example, a use
```
mounirlamouri: typo: in order

index.html
marcoscaceres, line 108: Sure, will split it up again. But do take a look after. It means you need to go to two different places to find essentially the same thing (definitions and links to definitions).

index.html
marcoscaceres: Hmm... I thought I did link to it. That's a bug.

index.html
marcoscaceres: fixed

index.html
marcoscaceres: I can add that, but which spec are you basing that on? 

you really don't need the "that MUST be supported as attribute solely by the <a>ScreenOrientation</a> object." The WebIDL already says that.

index.html
marcoscaceres: heh, I might have "fixed" the one I fixed! :tada:

index.html
marcoscaceres: fixed

index.html
marcoscaceres: fixed

index.html
marcoscaceres: Added natural

mounirlamouri:
Crap, GitHub doesn't save draft comments... :(

I have a few general comments.
- It seems that you usually don't linkify `user agent` but the document usually does. With your changes, it becomes fairly inconsistent. Could you keep the `<a>` around it? Or have a commit to remove all of them. I don't care much but I would like consistency :)
- It bothers me a bit that we return `SecurityError` for fullscreen and sandbox. It might make it hard to understand what's happening but also, the fullscreen requirement is more a UX issue than a security error. I don't see a better solution though.
- Why moved conformance at the bottom? It is usually at the beginning and I'm not sure what's the benefit of moving it at the bottom.

I also have a couple of comments from your initial list of changes:
- rewrote how to get attributes
  -> what did you do exactly?
- removed redundancy when invoking lock() and unlock()
  -> what did you do exactly?

Finally, could you add your name after mine in the author list?

index.html
mounirlamouri: http://xhr.spec.whatwg.org

index.html
mounirlamouri, line 108: I have an idea. What about adding the non-dependency things at the beginning of "Concepts".
Like:

```
<section>
<h2>Concepts</h2>
When the spec says <b>screen</b> it means unicorn.
For every top level browsing window, this specification has a pending kitten.
<section>
<h2>Reading the screen orientation</h2>
[...]
```

marcoscaceres, line 108: like it! :+1:

index.html
marcoscaceres: You've been reading **Van** Kesteren! I should have known!

marcoscaceres: Updated it to this:

``` HTML
          The <code><dfn id=
          "widl-ScreenOrientation-onchange">onchange</dfn></code> attribute is
          an <a>event handler</a> whose corresponding <a>event handler event type</a> 
          is <code>change</code>.
```

That matches what Anne's spec has, includes the name of the event handler event type, without having to repeat the information that is already in the WebIDL.

index.html
marcoscaceres: fixed

index.html
mounirlamouri: Sounds good. Make sure to add the &lt;dfn> in the Definitions section.

index.html
marcoscaceres: linked up all "user agents" now.

index.html
marcoscaceres: Yep, added definitions. Thanks for the reminder.

index.html
marcoscaceres, line 627: We should do this as a separate bug.

marcoscaceres, line 627: Created #28

index.html
marcoscaceres: Do you mean as part of the task?

marcoscaceres: Ok, yes. What I'm not sure of is if events fire before resolving the promise or after.

index.html
marcoscaceres: hmmm... ok, this might require a bit of work then. Need to somehow perform this comparison but I don't know if we have everything we need (as inputs) to do that in the algorithm.

index.html
mounirlamouri: I think you missed that comment.

kenchris:
@mounirlamouri comment in the files view instead and they won't disappear when force pushed

PR #35, opened 2014-07-18
Note explaining why unlock() does not return a Promise.

mounirlamouri:
@marcoscaceres ? :)


mounirlamouri:
This should fix #31

marcoscaceres:
Makes sense.

Issue #53, opened 2014-08-14
Promise<void> ?


marcoscaceres:
Elsewhere, @dontcallmedom wrote:

> - Promise`<undefined>` should be Promise`<void>` from my reading of WebIDL

marcoscaceres:
Need to ask public-script-coord. 
Essentially, this changes what `arguments.length` equals to. It would make sense to call it with no argument instead of `undefined` tho.

marcoscaceres:
Seems it should be `void` and resolving with `void` too.

domenic:
I wouldn't resolve with `void`, although I guess it would work... resolving with `undefined` seems clearer and also allowed?

marcoscaceres:
Argh, the IDL spec is mega confusing here. "undefined" is not really defined as a type. But then it says:

> Functions on platform objects that implement an operation whose IDL specifies a void return type MUST return the undefined value.

marcoscaceres:
cc @heycam.

heycam:
> Functions on platform objects that implement an operation whose IDL specifies a void return type MUST return the undefined value.

This is just saying that if you have say

```
interface A {
  void f();
};
```

then the JS `Function` object that implements `A.f` must return undefined.

As for using `Promise<void>`: Web IDL doesn't have definitions for "resolve promise with this value" or "reject promise with this value". Arguably it should, so that we can have consistent handling of ES -> IDL type conversions. If people write specs now with wording like that, then I think that's basically calling in to the definitions in the ES spec.

Web IDL _does_ have a definition called "perform some steps once a promise is settled", which is an abstraction over calling `then()` on the promise, and then doing appropriate type conversions from JS values to IDL values so that your prose can continue along with IDL values for whatever the Promise was resolved with. That's the only thing in the spec at the moment that does anything with the type parameter of `Promise<>`; if you have a `Promise<void>` object, then you can write prose that says for example "When the promise object is resolved, ..." where "..." doesn't have to deal with any particular value that the Promise was resolved with.  If you had a `Promise<DOMString>` or whatever, your "..." could refer to the value the promise was resolved with and know that it's a `DOMString` value.

mounirlamouri:
Should be fixed, please feel free to reopen if you disagree.

PR #55, opened 2014-08-15
Use Promise<void> instead of Promise<undefined>.

mounirlamouri:
Fixes issue #53.

@marcoscaceres, what do you think? Does that matches the conclusion of the discussions?


marcoscaceres:
Yep, pretty much.

PR #58, opened 2014-08-22
Use fullscreen element stack instead of fullscreen enabled flag.

### LINKED ISSUE #38: Fix references to HTML spec
[Elsewhere](http://lists.w3.org/Archives/Public/public-webapps/2014JulSep/0188.html), Paul Cotton noted that: 

> Section 6 [1] refers to bug 25310 [2] and suggests that the HTML5 sandboxing material could be changed. The current specification refers to HTML5.0 [3] which is currently in its last CR period before progressing to Proposed Recommendation. 
> If a change is required in the HTML specification then it is more likely that this would occur in HTML 5.1 [4]. 

I would strongly urge us to bypass all this and just reference the WHATWG HTML (i.e., [[!HTML]]).

mounirlamouri:
Fixes #38.

@annevk and @marcoscaceres WDYT?

### DIFF
--- a/index.html
+++ b/index.html
@@ -170,8 +170,8 @@
       </p>
       <p>
         <dfn><a href=
-        'http://fullscreen.spec.whatwg.org/#fullscreen-enabled-flag'>fullscreen
-        enabled flag</a></dfn> and <dfn><a href=
+        'http://fullscreen.spec.whatwg.org/#fullscreen-element-stack'>fullscreen
+        element stack</a></dfn> and <dfn><a href=
         'http://fullscreen.spec.whatwg.org/#fully-exit-fullscreen'>fully exits
         fullscreen</a></dfn> are defined in [[!FULLSCREEN]].
       </p>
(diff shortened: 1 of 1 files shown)


annevk:
Looks okay, though this does not cover the case where you would exit fullscreen through normal means.

index.html
```diff
@@ -728,11 +728,11 @@
...
-          the top-level browsing document's <a>fullscreen enabled flag</a> has
-          been set. When that security condition applies, whenever the document
-          <a>fully exits fullscreen</a> and a screen orientation lock is
-          applied, the <a>user agent</a> MUST lock the orientation to the
-          default orientation.
+          the <a>top-level browsing context</a>'s <a>document</a>'s
+          <a>fullscreen element stack</a> is not empty. When that security
+          condition applies, whenever the document <a>fully exits fullscreen</a>
+          and a screen orientation lock is applied, the <a>user agent</a> MUST
+          lock the orientation to the default orientation.
```
marcoscaceres: Add cross-ref to `<a>default orientation</a>`

mounirlamouri:
#annevk, is it better? I'm no longer using "fully exit fullscreen" but uses the fullscreen element stack being empty instead.

mounirlamouri:
@annevk said it was okay on IRC even though it would be nicer with actual hooks. As long as those hooks are not in the fullscreen spec, I doubt we can do bettter here. Merging.

PR #61, opened 2014-08-25
Improve readability regarding active/current orientation lock.

### LINKED ISSUE #46: Rename variables to more clearly indicate what is a list
Elsewhere, @annevk pointed out: 

> "If the active orientation lock is not the current orientation lock" 
> how does this work given that at least current orientation lock is a 
> list? (It would help to rename these variables to more clearly 
> indicate what is a list. That might also help avoid such errors.

mounirlamouri:
This commit is renaming "current orientation lock" as "document's orientation lock" and is clearer about the "active orientation lock" being one document's orientation lock.

It should hopefully fix #46

### DIFF
--- a/index.html
+++ b/index.html
@@ -373,8 +373,8 @@
           <a><code>OrientationLockType</code></a> enum.
         </p>
         <p>
-          The <dfn>current orientation lock</dfn> is the orientation lock that
-          applies on a <a>top-level browsing context</a>.
+          A <a>document</a>'s <dfn>orientation lock</dfn> is the orientation
+          lock that applies on its <a>top-level browsing context</a>.
         </p>
         <p>
           The steps to <dfn>apply an orientation lock</dfn> using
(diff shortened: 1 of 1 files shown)


mounirlamouri:
@annevk do you prefer that version?

mounirlamouri:
An orientation lock doesn't have to be one value. You can lock to multiple values like "portrait" might lock to "portrait-primary' and "portrait-secondary" but they will be seen as one orientation lock.

I agree that the save isn't great will rewrite that part.

mounirlamouri:
@annevk, please take another look.

annevk:
If a lock can have multiple values, why isn't it a list or a set as I suggested elsewhere?

mounirlamouri:
Anne, I think I applied all your comments (list + "is equivalent"). I'm merging this. Feel free to open another issue if there is something else you want me to change.

PR #67, opened 2014-09-10
Defines context for "apply an orientation lock" and "lock the orientation" (fixes #44)

### LINKED ISSUE #44: What's the context for "apply an orientation lock"
Elsewhere, @annevk demanded to know: 

> "apply an orientation lock" has no information about context. E.g. 
> what document are we talking about here? That could be more explicit. 
> That it has to be a top-level browsing context's document is also 
> unclear.

mounirlamouri:


index.html
```diff
@@ -132,6 +132,10 @@
           'http://www.w3.org/TR/html5/browsers.html#active-sandboxing-flag-set'>
           active sandboxing flag set</a></dfn>
         </li>
+        <li>
+          <dfn><a href=
+          'http://www.w3.org/TR/html5/webappapis.html#responsible-document'>
```
marcoscaceres, line 137: Should link to: 
http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html#responsible-document

mounirlamouri, line 137: I prefer to keep consistency and use w3.org/TR/HTML5 until we decide to move everything. With the current discussions regarding WHATWG dependencies in W3C docs, I would be careful with that though.

PR #68, opened 2014-09-10
Rewrite the updating and reading part of the specification.

mounirlamouri:
This is rewriting how we store the orientation information and how it is updated. It is moving the pending promise concept to the document and changes how and when we fire the change events. It also clarifies the fact that the value is stalled when the page is invisible.

Should fix all the open LC bugs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -369,12 +369,13 @@
           Locking the screen orientation
         </h2>
         <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Going through commits independently

PR #74, opened 2014-09-18
Use GitHub issues instead of W3C bugzilla.

mounirlamouri:


mounirlamouri:
@marcoscaceres FYI

marcoscaceres:
thanks!

marcoscaceres:
btw, do you want me to add something like we have for manifest at the top of the doc?

mounirlamouri:
How does that work with the W3C pubrules? Given that we are going for LC, I would prefer to not mess with the pubrules now. We could do after LC for the ED.

marcoscaceres:
I has effect - promise.

Issue #72, opened 2014-09-12
4.4 Handling screen orientation changes 


annevk:
- Both algorithms listed here lack detail about whether they run asynchronously or not. 
- They are unclear from what task they run.
- They do not queue tasks to synchronize but rather just dispatch events.
- You are monkey patching PAGE-VISIBILITY.

mounirlamouri:
> Both algorithms listed here lack detail about whether they run asynchronously or not.

Will fix that.

> They are unclear from what task they run.

Ideally, that should happen in the animation frame task source but that's not really defined yet and in the purpose of moving to LC, I don't think it's a good idea to depend on unclear concepts. I will add a comment to the animation frame task source issue that we have opened.

> They do not queue tasks to synchronize but rather just dispatch events.

Will fix that.

> You are monkey patching PAGE-VISIBILITY.

I assume you are referring to the second paragraph? To me, it's not really monkey patching, it's mostly a wait to say:

``` js
document.addEventListener('visibilitychange', function() {
  if (document.visibilityState == 'visible') {
    // steps
  }
}
```

annevk:
That is monkey patching... As e.g. if I implemented it that way it would be different if someone overwrote `document.visibilityState`.

As for animation frame task, you should point that out in the specification. It's a disservice to implementers to hide such essential details for the purposes of some silly process you're trying to go through.

mounirlamouri:
> That is monkey patching... As e.g. if I implemented it that way it would be different if someone
> overwrote document.visibilityState.

I will send a feedback to the web-perf mailing list. Can't really do more.

> As for animation frame task, you should point that out in the specification. It's a disservice to 
> implementers to hide such essential details for the purposes of some silly process you're trying to go 
> through.

I opened #75 to do that. Will work on this after the LC milestone.

mounirlamouri:
I opened an issue (#77) about improving how we hook into Page Visibility. I've also sent an email to the web-perf mailing list to request a better hook: http://lists.w3.org/Archives/Public/public-web-perf/2014Sep/0034.html

I will write a patch to use "now visible algorithm". Should be better.

mounirlamouri:
I think all the issues raised here have been fixed or have a follow-up issue.

PR #76, opened 2014-09-18
Apply comments from @annevk in #72.

mounirlamouri:
@annevk and @marcoscaceres how does that look?

I think it fixes the issues that are fixable in #72. I also removed the "cancel all tasks" because I just have no idea why that's there and it sounds pretty dangerous.


mounirlamouri:
I'm merging that with your comment applied. Feel free to leave a comment if you disagree with anything there.

PR #78, opened 2014-09-18
Use 'now visible algorithm' for the PAGE-VISIBILITY hook.

mounirlamouri:


mounirlamouri:
@annevk and @marcoscaceres WDYT?

mounirlamouri:
I think that would close #72.

mounirlamouri:
Per discussion on IRC, we will go with that.

PR #86, opened 2015-06-07
Use current orientation angle instead of screen orientation angle in normative text.

mounirlamouri:
This is fixing issue 85.


tobie:
:thumbsup:

PR #90, opened 2016-07-11
Add orientation change source to be used by fullscreen api.

mounirlamouri:
@foolip PTAL

### DIFF
--- a/index.html
+++ b/index.html
@@ -124,6 +124,11 @@
           event handler event type</a></dfn>
         </li>
         <li>
+          <dfn><a href=
+          'https://html.spec.whatwg.org/multipage/webappapis.html#concept-task'>
+          task</a></dfn>
+        </li>
+        <li>
           <code><dfn><a href=
           'https://html.spec.whatwg.org/multipage/browsers.html#window'>Window</a></dfn></code>
         </li>
@@ -653,6 +658,13 @@
                   </li>
                 </ol>
               </li>
+              <li>If the orientation change was triggered by a user gesture such
+              as the user turning the device, as opposed to a call to <a for=
+              'ScreenOrientation'>lock</a>, the <a>task</a> MUST be annotated
+              with <code>process user orientation change</code> when running the
+              next step.
+              </li>
+            </li>
               <li>
                 <a>Fire a simple event</a> named <code>change</code> at
                 <var>doc</var> 's <a href=
@@ -692,6 +704,11 @@
                   </li>
                 </ol>
               </li>
+              <li>If the orientation change was triggered by a user gesture such
+              as the user turning the device, as opposed to a call to <a for=
+              'ScreenOrientation'>lock</a>, the <a>task</a> MUST be annotated
+              with <code>process user orientation change</code> when running the
+              next step.</li>
               <li>
                 <a>Fire a simple event</a> named <code>change</code> at the
                 <a>document</a>'s <a href=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -701,6 +701,11 @@
             </ol>
           </li>
         </ol>
+        <p>
+          An <dfn>orientation change source</dfn> is said to be from a
```
foolip: What would the wording in the Fullscreen API be like? Paging @annevk to review the answer to that question :)

And won't `<code>gesture</code>` just make "gesture" monospaced as if it's the name of an event or similar?

Is it fair to say that the bits added here can _only_ be used in a way that guarantees a bad frame?

mounirlamouri:
> What would the wording in the Fullscreen API be like? Paging @annevk to review the answer to that question :)

Sent a PR.

> And won't <code>gesture</code> just make "gesture" monospaced as if it's the name of an event or similar?

Correct. I can change this to be a boolean answer if you prefer.

> Is it fair to say that the bits added here can only be used in a way that guarantees a bad frame?

I would say so, unless an implementation fires the event but holds rendering a new frame.

foolip:
OK, from https://github.com/whatwg/fullscreen/pull/43 it looks like "gesture" is a value that "orientation change source" can have, so it's bit like an enum with only one value.

I think that a cleaner split might be an algorithm similar to https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation in the Screen Orientation spec, perhaps called "triggered by user screen orientation change"?

mounirlamouri:
@foolip PTAL

mounirlamouri:
Changed a bit to move all the logic to this spec so fullscreen can just link to the concept without having to worry about any kind of logic.

annevk:
`screen.orientation` is a little vague. You need to get at that object somehow. Not sure what the current machinery is to get to a global/document from a task. @domenic?

domenic:
You need to save the appropriate global before going async, in general. Once you're in the event loop you've got no reference to such objects, given that an event loop can service multiple globals.

Off the top of my head, maybe one way to do this is to associate an orientation change counter with every event loop (you can do this from within the screen orientation spec), and increment it at the beginning of every "change" event firing, and decrement it at the end of every change event firing (after all handlers have run). Then you can check if the event loop's orientation change counter is greater than zero.

annevk:
So how does that work with user interaction events? They are triggered outside the context of an event loop and end up queuing a task to do some stuff. That task then needs to find out what is affected. I guess that can look at the current global when it runs?

domenic:
It's quite possible those events aren't well defined either. But no, they can't look up the current global when they run, because there is no current global (unless they go back and invoke JS code in some specific saved global, such as the global of a callback function).

mounirlamouri:
The "triggered by user activation" is defined very simply. It's only using the event type. It actually allows `change` so maybe there isn't much to do here ;) Maybe I could say "screen orientation `change` event"? My intent here was to be a bit more precise than juste `change` event because there are many.

annevk:
"triggered by user activation" might well be wrong too.

What probably happens in practice is that the "task" is annotated in some way. And if it's annotated in the way we want, stuff will work. For some user interaction activities screen orientation will need to queue tasks. It could annotate those tasks somehow and that's what we'd use. E.g., Fetch annotates some tasks as https://fetch.spec.whatwg.org/#process-request-body.

mounirlamouri:
@annevk @domenic @foolip PTAL. I think this should match the discussion above now.

domenic:
If the others are ok with this model, it seems reasonable to me. Eventually we should try to investigate how implementations actually do all this stuff, but that's a larger project.

annevk:
LGTM as a first take.

PR #108, opened 2017-11-01
Fix all ReSpec warnings

alexshalamov:
This PR fixes ReSpec warnings and improves cross-linking inside document.

### DIFF
--- a/index.html
+++ b/index.html
@@ -223,7 +223,7 @@
       </section>
       <section>
         <h2>
-          <code>ScreenOrientation</code> interface
+          <code><dfn>ScreenOrientation</dfn></code> interface
         </h2>
         <pre class='idl'>
           [Exposed=Window]
@@ -248,7 +248,7 @@
           <a>responsible document</a>'s <a>default orientation</a>.
         </p>
         <p class='note'>
-          <a for='ScreenOrientation' data-lt='unlock'>unlock()</a> does not
+          <a data-link-for='ScreenOrientation' data-lt='unlock'>unlock()</a> does not
           return a Promise because it is equivalent to locking to the
           <a>default orientation</a> which might or might not be known by the
           <a>user agent</a>. Hence, the <a>user agent</a> can not predict what
@@ -274,7 +274,7 @@
       </section>
       <section>
         <h2>
-          <code>OrientationType</code> enum
+          <code><dfn>OrientationType</dfn></code> enum
         </h2>
         <pre class='idl'>
           enum OrientationType {
@@ -287,7 +287,7 @@
       </section>
       <section>
         <h2>
-          <code>OrientationLockType</code> enum
+          <code><dfn>OrientationLockType</dfn></code> enum
         </h2>
         <pre class='idl'>
           enum OrientationLockType {
@@ -360,7 +360,7 @@
         <ol>
           <li>If the screen width is greater than the screen height, set the
           <a>document</a>'s <a>current orientation type</a> to
-            <code><dfn for='OrientationTypelandscape'>landscape-primary</dfn></code>
+            <code><dfn for='OrientationType'>landscape-primary</dfn></code>
             or <code><dfn for=
             'OrientationType'>landscape-secondary</dfn></code>.
           </li>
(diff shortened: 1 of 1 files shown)


alexshalamov:
@marcoscaceres please take a look.

index.html
```diff
@@ -236,6 +236,13 @@
           };
         </pre>
         <p>
+          The <dfn>ScreenOrientation</dfn> interface provides functionality to
+          <code><a data-link-for='ScreenOrientation' data-lt='lock'>lock()</a>
+          </code> and <code><a data-link-for='ScreenOrientation' data-lt=
+          'unlock'>unlock()</a></code> screen orientation, as well as to get
+          the orientation information about the <a>Screen</a> object.
+        </p>
+        <p>
```
mounirlamouri: Why did you add this?

alexshalamov: Because definition for `ScreenOrientation` interface was missing.

alexshalamov: I can remove this section and add `<dfn>` to the header of the section.

mounirlamouri: I haven't used respec in a while but maybe you could simply replace the `<a>ScreenOrientation</a>` with `<code>ScreenOrientation</code>`?

alexshalamov: Fixed. Thanks

mounirlamouri (approved):

mounirlamouri:
Thanks! :)

index.html
marcoscaceres: No, don't ever use code :( Only use`<a>` or it won't auto link.

mounirlamouri: dfn ended up being used but in the title instead of adding a new paragraph. Though, I wonder why the interface in the WebIDL snippet wasn't linkified.

index.html
```diff
@@ -223,7 +223,7 @@
       </section>
       <section>
         <h2>
-          <code>ScreenOrientation</code> interface
+          <code><dfn>ScreenOrientation</dfn></code> interface
```
marcoscaceres, line 226: you don't need the code elements here.

index.html
```diff
@@ -248,7 +248,7 @@
           <a>responsible document</a>'s <a>default orientation</a>.
         </p>
         <p class='note'>
-          <a for='ScreenOrientation' data-lt='unlock'>unlock()</a> does not
+          <a data-link-for='ScreenOrientation' data-lt='unlock'>unlock()</a> does not
```
marcoscaceres, line 251: we should move `data-link` up to the section element.

index.html
```diff
@@ -274,7 +274,7 @@
       </section>
       <section>
         <h2>
-          <code>OrientationType</code> enum
+          <code><dfn>OrientationType</dfn></code> enum
```
marcoscaceres, line 277: same here about `code`

marcoscaceres: ... and globally.

index.html
```diff
@@ -360,7 +360,7 @@
         <ol>
           <li>If the screen width is greater than the screen height, set the
           <a>document</a>'s <a>current orientation type</a> to
-            <code><dfn for='OrientationTypelandscape'>landscape-primary</dfn></code>
+            <code><dfn for='OrientationType'>landscape-primary</dfn></code>
```
marcoscaceres, line 363: please remove the for attribute, as it's not supported.

index.html
```diff
@@ -662,12 +662,11 @@
                 </ol>
               </li>
               <li>If the orientation change was triggered by a user gesture such
-              as the user turning the device, as opposed to a call to <a for=
-              'ScreenOrientation'>lock</a>, the <a>task</a> MUST be annotated
-              with <code>process user orientation change</code> when running the
-              next step.
+              as the user turning the device, as opposed to a call to
+              <a data-link-for='ScreenOrientation' data-lt='lock'>lock</a>, the
```
marcoscaceres, line 666: this will auto link once you move `data-link-for` up to the section element.

index.html
alexshalamov: @marcoscaceres I tried not to remove anything that was already in the spec. Should I do it in follow-up PR or you want to revert this one?

index.html
alexshalamov: @marcoscaceres Thanks for the hint, can I make new PR for these fixes?

index.html
marcoscaceres: A new new PR would be great. Really appreciate your time and help here.

PR #109, opened 2017-11-02
Clean-up anchors and definitions

alexshalamov:
Based on feedback from Marcos, this PR introduces following changes:

- Removes unnecessary `<code></code>` tags around definitions
- Adds data-dfn-for and data-link-for for sections to avoid duplication
- Replaces 'for' attribute with 'data-dfn-for'

### DIFF
--- a/index.html
+++ b/index.html
@@ -106,87 +106,73 @@
       </p>
       <ul>
         <li>
-          <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task">queue a
-          task</a></dfn>
+          <dfn data-cite="!HTML/webappapis.html#queue-a-task">queue a
+          task</dfn>
         </li>
         <li>
-          <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/webappapis.html#event-handlers">event
-          handler</a></dfn>
+          <dfn data-cite="!HTML/webappapis.html#event-handlers">event
+          handler</dfn>
         </li>
         <li>
-          <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-event-type">
-          event handler event type</a></dfn>
+          <dfn data-cite="!HTML/webappapis.html#event-handler-event-type">
+          event handler event type</dfn>
         </li>
         <li>
-          <dfn><a href=
-          'https://html.spec.whatwg.org/multipage/webappapis.html#concept-task'>
-          task</a></dfn>
+          <dfn data-cite="!HTML/webappapis.html#concept-task">
+          task</dfn>
         </li>
         <li>
-          <dfn><a href=
-          'https://html.spec.whatwg.org/multipage/browsers.html#window'>Window</a></dfn>
+          <code><dfn data-cite="!HTML/window-object.html#window">Window</dfn>
+          </code>
         </li>
         <li>
-          <dfn><a href=
-          'https://html.spec.whatwg.org/multipage/dom.html#document'>Document</a></dfn>
+          <code><dfn data-cite="!HTML/dom.html#document">Document</dfn></code>
         </li>
         <li>
-          <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/browsers.html#browsing-context">browsing
-          context</a></dfn>
+          <dfn data-cite="!HTML/browsers.html#browsing-context">browsing
+          context</dfn>
         </li>
         <li>
-          <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context">
-          top-level browsing context</a></dfn>
+          <dfn data-cite="!HTML/browsers.html#top-level-browsing-context">
+          top-level browsing context</dfn>
         </li>
-        <li>browsing context's <dfn><a href=
-        'https://html.spec.whatwg.org/multipage/browsers.html#active-document'>active
-        document</a></dfn>
+        <li>browsing context's <dfn data-cite=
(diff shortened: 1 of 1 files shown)


alexshalamov:
@marcoscaceres this PR addresses comments that you gave in #108

index.html
```diff
@@ -131,12 +131,12 @@
           task</a></dfn>
         </li>
         <li>
-          <code><dfn><a href=
-          'https://html.spec.whatwg.org/multipage/browsers.html#window'>Window</a></dfn></code>
+          <dfn><a href=
```
marcoscaceres: We need to update these (I mislead you a bit, here is ok to use code, then ReSpec automatically "codes" any `a` that xref this)... we can do these separately tho unless you feel like doing them now. 

```HTML
<code><dfn data-cite="!HTML/multipage/browsers.html#window">Window</dfn></code>
```

See: https://github.com/w3c/respec/wiki/data--cite

index.html
```diff
@@ -231,45 +231,44 @@
           };
         </pre>
         <p>
-          When the <code><dfn for='ScreenOrientation'>lock</dfn>()</code>
-          method is invoked, the <a>user agent</a> MUST run the <a>apply an
-          orientation lock</a> steps to the <a>responsible document</a> using
+          When the <dfn>lock</dfn>() method is invoked,
```
marcoscaceres: Nit: we now support `<dfn>lock()</dfn>` (i.e., parens inside)

index.html
```diff
@@ -212,13 +212,13 @@
           };
         </pre>
         <p>
-          The <code><dfn for='Screen'>orientation</dfn></code> object is an
+          The <dfn for='Screen'>orientation</dfn> object is an
```
marcoscaceres: Nit: for attribute is not supported.

index.html
```diff
@@ -231,45 +231,44 @@
...
+          When the <dfn>lock</dfn>() method is invoked,
+          the <a>user agent</a> MUST run the <a>apply an orientation lock</a>
+          steps to the <a>responsible document</a> using
           <var>orientation</var>.
         </p>
         <p>
-          When the <code><dfn for='ScreenOrientation'>unlock</dfn>()</code>
-          method is invoked, the <a>user agent</a> MUST run the steps to
-          <a>lock the orientation</a> of the <a>responsible document</a> to the
+          When the <dfn>unlock</dfn>() method is
```
marcoscaceres: as above.

index.html
```diff
@@ -231,45 +231,44 @@
...
+          the <a>user agent</a> MUST return the <a>responsible document</a>'s
+          <a>current orientation angle</a>.
         </p>
         <p>
-          The <code><dfn for='ScreenOrientation'>onchange</dfn></code>
-          attribute is an <a>event handler</a> whose corresponding <a>event
-          handler event type</a> is <code>change</code>.
+          The <dfn>onchange</dfn> attribute is an
+          <a>event handler</a> whose corresponding <a>event handler event type</a>
+          is <code>change</code>.
```
marcoscaceres: we should probably write `<code>"change"</code>`... if we defined "change", we should probably link to it... can't remember tho.

index.html
```diff
@@ -231,45 +231,44 @@
...
         </p>
       </section>
       <section>
```
marcoscaceres: Please add `data-link-for="OrientationType" data-dfn-for="OrientationType"` attributes to this section.

index.html
```diff
@@ -355,15 +354,13 @@
         <ol>
           <li>If the screen width is greater than the screen height, set the
           <a>document</a>'s <a>current orientation type</a> to
-            <code><dfn for='OrientationType'>landscape-primary</dfn></code>
-            or <code><dfn for=
-            'OrientationType'>landscape-secondary</dfn></code>.
+          <dfn data-dfn-for='OrientationType'>landscape-primary</dfn> or
```
marcoscaceres: Let's move the `data-dfn-for='OrientationType'` to the `section` element, then it will just work with `<dfn>`

index.html
```diff
@@ -463,20 +460,20 @@
...
               <dt>
-                <code><dfn for=
-                'OrientationLockType'>portrait-primary</dfn></code> or
-                <code><dfn for=
-                'OrientationLockType'>portrait-secondary</dfn></code> or
-                <code><dfn for=
-                'OrientationLockType'>landscape-primary</dfn></code> or
-                <code><dfn for=
-                'OrientationLockType'>landscape-secondary</dfn></code>
+                <dfn data-dfn-for=
```
marcoscaceres: as above... then you don't need `data-dfn-for` here... ReSpec always looks for the nearest section with "data-dfn-for=".

index.html
```diff
@@ -702,9 +698,9 @@
               </li>
               <li>If the orientation change was triggered by a user gesture
               such as the user turning the device, as opposed to a call to
-              <a data-link-for='ScreenOrientation'>lock</a>, the <a>task</a>
-              MUST be annotated with <code>process user orientation change
-              </code> when running the next step.
+              <a>lock</a>, the <a>task</a> MUST be annotated with
+              <code>process user orientation change</code> when running the
```
marcoscaceres, line 684: Nit: this `code` should probably have been an `<a>`

marcoscaceres (changes requested):
Nearly there!

index.html
alexshalamov: Fixed all comments except this one, as it needs definition or algo for `process user orientation change` term.

marcoscaceres (approved):
Thanks again @alexshalamov for your work on this!

PR #110, opened 2017-11-02
Reference correct promise at the "apply an orientation lock" operation

### LINKED ISSUE #94: Confusiong between "_pending-promise_" and "doc's pending promise"
In step 6 of "apply an orientation lock", suddenly the variable _pending-promise_ appears. It gets set to something before it is ever initialized. And each document's pending promise is only ever set to null, and never set to a promise.

I think all references to _pending-promise_ are supposed to be to an appropriate document's pending promise.

pozdnyakov:
Fixes #94. In the "apply an orientation lock" operation reference
the doc's pending promise instead of the 'pending-promise' variable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -512,9 +512,9 @@
               </dd>
             </dl>
           </li>
-          <li>Set <var>pending-promise</var> to be a newly-create a promise.
+          <li>Set <var>doc</var>'s <a>pending promise</a> to a newly-created promise.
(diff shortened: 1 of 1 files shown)


pozdnyakov:
@marcoscaceres @mounirlamouri PTAL

marcoscaceres (approved):

PR #111, opened 2017-11-02
Add [SameObject] to the `Screen.orientation` attribute

pozdnyakov:
There should not be any new Orientation instances created on a Screen.orientation call.


mounirlamouri:
lgtm, I will let @marcoscaceres have the last word though.

marcoscaceres (approved):

PR #112, opened 2017-11-03
Fix event firing

### LINKED ISSUE #93: Event firing needs fixes
"Fire a simple event" has been removed. Instead just use [fire an event](https://dom.spec.whatwg.org/#concept-event-fire).

Also, very importantly, since you are firing the events from asynchronous browser code, you need to specify which constructor you are using. So e.g.

> "[Fire an event] named `change` at _doc_'s `screen.orientation` object, using the `Event` constructor from _doc_'s [relevant Realm]"

/cc @annevk @bzbarsky

pozdnyakov:
Fixes #93. Use "fire an event" operation instead of the removed "fire a simple event".

### DIFF
--- a/index.html
+++ b/index.html
@@ -112,11 +112,6 @@
         </li>
         <li>
           <dfn><a href=
-          "https://html.spec.whatwg.org/multipage/webappapis.html#fire-a-simple-event">fire
-          a simple event</a></dfn>
-        </li>
-        <li>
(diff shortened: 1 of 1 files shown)


pozdnyakov:
@marcoscaceres @mounirlamouri PTAL

mounirlamouri:
lgtm

PR #114, opened 2018-05-08
Clarify the behavior of ScreenOrientation.angle

### LINKED ISSUE #79: Clarify orientation.angle behavior
@sicking wrote: 

> Also, I can't find any normative definition of if orientation.angle 
> should increase or decrease if the user rotates a device 90 degrees 
> clockwise? 
> 
> My bad, I see it now. Given how easy this is to get wrong, it might be 
> worth adding this information more explicitly at the definition of the 
> 'angle' property, or the 'current orientation angle' concept, rather 
> than just buried inside an algorithm.

xfq:
Fix #79.

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,6 +245,19 @@
           MUST return the <a>responsible document</a>'s <a>current orientation
           angle</a>.
         </p>
+        <div class='note'>
+            <p><a>angle</a> indicates how far the user has turned the device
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
LGTM, @mounirlamouri - ok from you?

mounirlamouri:
lgtm

PR #115, opened 2018-05-08
Add Privacy and Security Considerations section

### LINKED ISSUE #96: Lacks a privacy and security consideration section
Please include information from the security and privacy self-review, even if it's an explanation why you see no privacy issues.

xfq:
Fix #96.

-----

One question I haven't answered yet is:

* How should this specification work in the context of a user agent’s "incognito" mode?

I don't know how implementations behave in this case.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/115.html" title="Last updated on Oct 12, 2018, 2:12 AM GMT (9e4ec29)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/115/2a16e96...9e4ec29.html" title="Last updated on Oct 12, 2018, 2:12 AM GMT (9e4ec29)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -890,9 +890,9 @@
         <h3>Access to aspects of a user’s local computing environment</h3>
         <p>
           The screen orientation type and angle of the device can be accessed with the API specified in this document,
-          and can be a potential fingerprinting vector. However using this information is very difficult compared to other fingerprinting information. 
+          and can be a potential fingerprinting vector. However, using this information is very difficult compared to other fingerprinting information. 
           Aggregating information derived from this API is not very helpful in fingerprinting. 
-          In particular the Device Orientation API provides a far more detailed version of the same information. 
+          In particular the <a href="https://w3c.github.io/deviceorientation/spec-source-orientation.html">DeviceOrientation Event Specification</a> provides a far more detailed version of the same information. 
           So in practice, this API's minimal fingerprinting potential is unlikely to be used by any competent attack.
         </p>
       </section>
@@ -902,8 +902,7 @@
         Acknowledgments
       </h2>
       <p>
-        Thanks to Marcos Cáceres, Christophe Dumez, Anne van Kesteren and
-        Chundong Wang, Fuqiao Xue for their useful comments.
+        Thanks to Marcos Cáceres, Christophe Dumez, Anne van Kesteren, Chundong Wang, Fuqiao Xue, and Chaals McCathie Nevile for their useful comments.
       </p>
       <p>
         Special thanks to Chris Jones and Jonas Sicking for their contributions
--- a/index.html
+++ b/index.html
@@ -885,17 +885,15 @@
         Privacy and Security Considerations
       </h2>
 
+     
       <section>
-        <h2>Access to sensors on a user’s device</h2>
+        <h3>Access to aspects of a user’s local computing environment</h3>
         <p>
-          This specification does not allow an origin access to sensors on a user’s device directly, but information exposed by sensors (e.g., the screen orientation angle) can be accessed indirectly.
-        </p>
-      </section>
-
-      <section>
-        <h2>Access to aspects of a user’s local computing environment</h2>
-        <p>
-          The screen orientation type and angle of the device can be accessed with the API specified in this document, and can be a potential fingerprinting vector.
+          The screen orientation type and angle of the device can be accessed with the API specified in this document,
+          and can be a potential fingerprinting vector. However using this information is very difficult compared to other fingerprinting information. 
+          Aggregating information derived from this API is not very helpful in fingerprinting. 
+          In particular the Device Orientation API provides a far more detailed version of the same information. 
+          So in practice, this API's minimal fingerprinting potential is unlikely to be used by any competent attack.
         </p>
       </section>
     </section>
@@ -905,7 +903,7 @@
       </h2>
       <p>
         Thanks to Marcos Cáceres, Christophe Dumez, Anne van Kesteren and
-        Chundong Wang for their useful comments.
+        Chundong Wang, Fuqiao Xue for their useful comments.
       </p>
       <p>
         Special thanks to Chris Jones and Jonas Sicking for their contributions
--- a/index.html
+++ b/index.html
@@ -880,6 +880,67 @@
 
 </pre>
     </section>
+    <section class='informative'>
+      <h2>
+        Privacy and Security Considerations
+      </h2>
+      <p>
+          Answers to the <a href="http://www.w3.org/2001/tag/">TAG</a>'s <a href="https://w3ctag.github.io/security-questionnaire/">self-review questionaire</a> are provided below.
+      </p>
+
+      <dl>
+      <dt>Does this specification deal with personally-identifiable information?
+      <dd>No.
+
+      <dt>Does this specification deal with high-value data?
+      <dd>No.
+
+      <dt>Does this specification introduce new state for an origin that persists across browsing sessions?
+      <dd>No.
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -880,6 +880,67 @@ <h2>
 
 </pre>
     </section>
+    <section class='informative'>
+      <h2>
+        Privacy and Security Considerations
+      </h2>
+      <p>
```
marcoscaceres: So, I tried doing exactly this format in the past (in a different spec) and people complained. Maybe we can just chop this down to the "yes" stuff, and make a paragraph out of it.

chaals: It's pretty useful to have the positive and negative aspects covered, but I agree that it should be readable prose rather than answers to questions whether they are relevant or not...

chaals (changes requested):
This answers the questions posed by the security/privacy questionnaire, but doesn't give much useful information.

I think the point is to explain what someone might do with that information that has an impact on security and privacy.

Although it is potentially possible to lock the orientation of a device, about the only application I can think of where there is an issue is something like a compass, where it is possible to actively mislead the user. In practice, that is already possible with a compass application.

Kowing whether the user is turning their device is indeed a potential fingerprinting vector (like so many things related to sensors, also because it can be corroborated externally, e.g. by processing surveillance video of an area). However, it is a relatively complex and inefficient attack to use - in practice I would not expect to see it deployed unless users become far-more privacy concsious and protect themselves against many simpler attacks.

index.html
```diff
@@ -895,14 +895,7 @@ <h2>Access to sensors on a user’s device</h2>
...
         <p>
-          The screen orientation type and angle of the device can be accessed with the API specified in this document.
-        </p>
-      </section>
-
-      <section>
-        <h2>Control over a user agent’s native UI</h2>
-        <p>
-          This specification allows locking the screen orientation to a specific state.
+          The screen orientation type and angle of the device can be accessed with the API specified in this document, and can be a potential fingerprinting vector.
```
chaals: I think it is worth saying how weak this **potential** vector is, and that it is unlikely to be used while far stronger ones are available.

IMHO, even trying to use this in association with other data is difficult, except real-time comparison of external visual surveillance or device-orientation.

For example, adding

> Using this information for practical fingerprinting is very difficult compared to other fingerprinting information. Aggregating information derived from this API is not very helpful in fingerprinting. The Device Orientation API provides a far more detailed version of the same information. So in practice, this API's minimal fingerprinting potential is unlikely to be used by any competent attack.

chaals: (I'll do that as a PR to your branch)

xfq:
I'll merge this in one day if there's no objection. Further improvements can be made later :)

marcoscaceres:
(need a few days still to review, sorry!)

index.html
```diff
@@ -880,13 +880,29 @@ <h2>
...
+      <h2>
+        Privacy and Security Considerations
+      </h2>
+
+     
+      <section>
+        <h3>Access to aspects of a user’s local computing environment</h3>
+        <p>
+          The screen orientation type and angle of the device can be accessed with the API specified in this document,
+          and can be a potential fingerprinting vector. However, using this information is very difficult compared to other fingerprinting information.
```
marcoscaceres: Please drop: "However, using this information is very difficult compared to other fingerprinting information."

index.html
```diff
@@ -880,13 +880,29 @@ <h2>
...
+        Privacy and Security Considerations
+      </h2>
+
+     
+      <section>
+        <h3>Access to aspects of a user’s local computing environment</h3>
+        <p>
+          The screen orientation type and angle of the device can be accessed with the API specified in this document,
+          and can be a potential fingerprinting vector. However, using this information is very difficult compared to other fingerprinting information. 
+          Aggregating information derived from this API is not very helpful in fingerprinting.
```
marcoscaceres: Same with this line.

index.html
```diff
@@ -880,13 +880,29 @@ <h2>
...
+      </h2>
+
+     
+      <section>
+        <h3>Access to aspects of a user’s local computing environment</h3>
+        <p>
+          The screen orientation type and angle of the device can be accessed with the API specified in this document,
+          and can be a potential fingerprinting vector. However, using this information is very difficult compared to other fingerprinting information. 
+          Aggregating information derived from this API is not very helpful in fingerprinting. 
+          In particular the <a href="https://w3c.github.io/deviceorientation/spec-source-orientation.html">DeviceOrientation Event Specification</a> provides a far more detailed version of the same information.
```
marcoscaceres: Same, all this stuff can go. Blaming another spec for being more leaky doesn't help this one :)

marcoscaceres (changes requested):
Making this section minimal would be better.

xfq:
@chaals Any comment on Marcos's comments? Maybe we need some research papers about this topic (see [Accelerometer](https://w3c.github.io/accelerometer/#security-and-privacy) for an example).

In the meantime, I also see [fingerprintjs2](https://github.com/Valve/fingerprintjs2) detecting the screen orientation to ensure consistent fingerprints when users rotate their mobile devices, although it doesn't use the Screen Orientation API: https://github.com/Valve/fingerprintjs2/blob/f6c39229228831ffb7d84592e0529633f25c0c65/fingerprint2.js#L164

index.html
```diff
@@ -880,13 +880,29 @@ <h2>
...
+     
+      <section>
+        <h3>Access to aspects of a user’s local computing environment</h3>
+        <p>
+          The screen orientation type and angle of the device can be accessed with the API specified in this document,
+          and can be a potential fingerprinting vector. However, using this information is very difficult compared to other fingerprinting information. 
+          Aggregating information derived from this API is not very helpful in fingerprinting. 
+          In particular the <a href="https://w3c.github.io/deviceorientation/spec-source-orientation.html">DeviceOrientation Event Specification</a> provides a far more detailed version of the same information. 
+          So in practice, this API's minimal fingerprinting potential is unlikely to be used by any competent attack.
+        </p>
```
mounirlamouri, line 901: I think we can say more than that. The spec provides the following information:
- angle
- type

Type is landscape / portrait which the website can already figure out with the `screen.width` and `screen.height`. The only information the `type` provides that wasn't available is the distinction between `primary` and `secondary`. It may be interesting to look deeper into these two sub-types.

Angle, by definition has a strong correlation with the `type` value. I wonder if it's worth looking more into it and how it may impact fingerprinting.

I don't think it's correct to say that this API is a sub-set of Device Orientation as Device Orientation, even if more noisy doesn't quite offer the same information.

marcoscaceres:
Closed by #121 -  @Johanna-hub did the change suggest here. Text is now in the spec. Thanks everyone for your input!

xfq:
Hmm, text is not in the spec now, because Johanna's PR was created against the `xfq/priv-sec` branch, which is the branch in this PR.

We should either merge both PRs, or rebase #121 against `gh-pages`...

Reopening this PR; Merging.

marcoscaceres:
> Hmm, text is not in the spec now, because Johanna's PR was created against the xfq/priv-sec branch, which is the branch in this PR.

Oh whoops, I should have spotted that.

marcoscaceres:
Thanks @xfq!

Johanna-hub:
Sorry about the confusion @marcoscaceres and @xfq and thanks for fixing. I created my PR against `xfq/priv-sec` so reviewers could evaluate the changes I made to the original PR. Should I have just used `gh-pages` branch?

marcoscaceres:
Either would be fine. Just note in a comment which branch it’s going into as it’s easy to miss.

PR #117, opened 2018-05-31
Add interaction with WCAG 2.1

xfq:
See https://lists.w3.org/Archives/Public/public-webapps/2018AprJun/0035.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -834,17 +834,18 @@
         </p>
 
         <p>
-            ​The intent of this Success Criterion is to ensure that all content and functionality is available regardless of the display
-            orientation (portrait or landscape). Some websites and applications automatically set the screen to a particular display
-            orientation and expect that users will respond by rotating their device to match.
+          ​The intent of this Success Criterion is to ensure that all
+          <a href="https://www.w3.org/TR/WCAG21/#dfn-essential">essential</a> content and functionality is available regardless of the display orientation (portrait or landscape). Some
+          websites and applications automatically set the screen to a particular display orientation and expect that users will respond
+          by rotating their device to match.
         </p>
 
         <p>
-            However, some users may have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore,
-            websites and applications need to support both orientations by making sure content and functionality is available in
-            each orientation. While the order of content and method of functionality may have differences the content and functionality
-            must always be available. When a particular orientation is essential, the user needs to be advised of the orientation
-            requirements.​
+          However, some users may have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore,
+          websites and applications need to support both orientations by making sure
+          <a href="https://www.w3.org/TR/WCAG21/#dfn-essential">essential</a> content and functionality is available in each orientation. While the order of content and method of functionality
+          may have differences the content and functionality must always be available. When a particular orientation is
+          <a href="https://www.w3.org/TR/WCAG21/#dfn-essential">essential</a>, the user needs to be advised of the orientation requirements.​
         </p>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -825,6 +825,28 @@
           for a web page using CSS.
         </p>
       </section>
+      <section class='informative'>
+        <h2>
+          Interaction with Web Content Accessibility Guidelines
+        </h2>
+        <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -825,6 +825,28 @@ <h2>
...
+
+        <p>
+            ​The intent of this Success Criterion is to ensure that all content and functionality is available regardless of the display
+            orientation (portrait or landscape). Some websites and applications automatically set the screen to a particular display
+            orientation and expect that users will respond by rotating their device to match.
+        </p>
+
+        <p>
+            However, some users may have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore,
+            websites and applications need to support both orientations by making sure content and functionality is available in
```
marcoscaceres: "need to support" is somewhat strong. Developers need to consider if the orientation lock is "[essential](https://www.w3.org/TR/WCAG21/#dfn-essential)" for the content.

marcoscaceres: ... and they should be mindful not to set an orientation unless it's absolutely [essential](https://www.w3.org/TR/WCAG21/#dfn-essential).

index.html
```diff
@@ -825,6 +825,28 @@ <h2>
...
+            each orientation. While the order of content and method of functionality may have differences the content and functionality
+            must always be available. When a particular orientation is essential, the user needs to be advised of the orientation
```
marcoscaceres: > While the order of content and method of functionality may have differences the content and functionality must always be available. 

I don't quite understand what this means. 

> When a particular orientation is essential, the user needs to be advised of the orientation ... 

It's a bit preachy... maybe:

When a particular orientation is essential, developers should inform users of the orientation requirement.

index.html
```diff
@@ -825,6 +825,28 @@ <h2>
...
+      <section class='informative'>
+        <h2>
+          Interaction with Web Content Accessibility Guidelines
+        </h2>
+        <p>
+            The Web Content Accessibility Guidelines 2.1 specification [[WCAG21]] includes a Success Criterion (<a href="https://www.w3.org/TR/WCAG21/#orientation">SC 1.3.4</a>) related to screen orientation.
+        </p>
+
+        <p>
+            ​The intent of this Success Criterion is to ensure that all content and functionality is available regardless of the display
```
LJWatson: Suggest adding "essential" into this sentence: "...intent of this Success Criterion is to ensure that all essential content and functionality..."

index.html
LJWatson: Also add "essential" into this sentence: "...Add line comment
+            However, some users may have their devices mounted in a fixed orientation (e.g. on the arm of a power wheelchair). Therefore,  
844
Add line comment
+            websites and applications need to support both orientations by making sure essential content and functionality is..."

marcoscaceres: Thanks @LJWatson for the additional input!

index.html
johnfoliot: >> While the order of content and method of functionality may have differences the content and functionality must always be available.
> I don't quite understand what this means.

It is a recognition that with CSS (i.e. flexbox, media queries) layout and content rendering may change when the orientation changes, however all content and functionality must be preserved. This may seem obvious on the surface, however... :) 

>> When a particular orientation is essential, the user needs to be advised of the orientation ...
> It's a bit preachy... maybe:
> When a particular orientation is essential, developers should inform users of the orientation requirement.

This is essentially an RFC 2119 MUST requirement. Replacing 'should' with 'must' would be preferred, however it is recognized that this is non-normative text in this Specification, so defer to Editors judgement.

index.html
johnfoliot: Hi @LJWatson ,

Essential is an undefined term here yet critical to the WCAG 2.1 spec, and I would be concerned about who determines what is essential. WCAG 2.1 has defined this term w.r.t. our spec, should this draft care to reference it: https://www.w3.org/TR/WCAG21/#dfn-essential

Again, it is recognized that this is non-normative text in this Specification, so defer to Editors judgement.

index.html
xfq: I kinda agree, but I can't think of better wording currently. Do you have any proposed text?

index.html
xfq: I think we usually try to avoid RFC 2119 key words in non-normative sections...

marcoscaceres:
@xfq ok to merge this, I think. I'll tidy up after.

xfq:
Thanks for your review! Merging.

PR #126, opened 2018-12-19
Edits Primary and Secondary orientation definitions

### LINKED ISSUE #113: Where are terms 'primary' and 'secondary' defined within spec?
It may be a fault in my education that I do not know what 'primary' or 'secondary' are supposed to mean here, but the specification is not helping by defining these terms.  Somehow the statement
>   However, the user agent can associate *-primary and *-secondary values at will.

is insufficient illumination.

Johanna-hub:
Fixes #113 

For feedback and discussion 🙂

Edits and clarifies 4.1 Reading the Screen Orientation according to the discussion in #113 to provide a better explanation of primary and secondary orientation.

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/126.html" title="Last updated on Jan 8, 2019, 2:54 PM UTC (c42df9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/126/f0e7a2e...Johanna-hub:c42df9f.html" title="Last updated on Jan 8, 2019, 2:54 PM UTC (c42df9f)">Diff</a>


index.html
```diff
@@ -297,11 +297,43 @@ <h2>
           For a given <a>document</a>, the <a>current orientation type</a> and
           the <a>current orientation angle</a> are strongly linked in the sense
           that for any given type, there will be a specific angle associated.
-          However, the <a>user agent</a> can associate <code>*-primary</code>
-          and <code>*-secondary</code> values at will. For example, if
-          <code>90</code> is associated with <code>landscape-primary</code> and
-          <code>270</code> with <code>landscape-secondary</code> for one
-          <a>document</a> , another one MAY get the opposite relationship.
+          However, the <a>user agent</a> can associate certain
+          <code>*-primary</code> and <code>*-secondary</code> values at will.
```
```suggestion
          <code>*-primary</code> and <code>*-secondary</code> values at will:
```

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          One primary orientation will always be determined by the natural
+          orientation of the device and this will then determine the secondary
+          value of its related orientation. For example a phone held in its
```
marcoscaceres: Breaking this up at the example might be good. This paragraph is getting rather long, and the example is non-normative. 

```suggestion
          value of its related orientation. </p> <p>For example, a phone held in its
```

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          natural orientation would have a current orientation of
+          <a>portrait-primary</a> and its <a>portrait-secondary</a> orientation
+          would be its position when rotated 180°. The decision on setting the
```
```suggestion
          would be its position when rotated 180°. </p> <p>The decision on setting the
```

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
           For a given <a>document</a>, the <a>current orientation type</a> and
           the <a>current orientation angle</a> are strongly linked in the sense
           that for any given type, there will be a specific angle associated.
-          However, the <a>user agent</a> can associate <code>*-primary</code>
-          and <code>*-secondary</code> values at will. For example, if
-          <code>90</code> is associated with <code>landscape-primary</code> and
-          <code>270</code> with <code>landscape-secondary</code> for one
-          <a>document</a> , another one MAY get the opposite relationship.
+          However, the <a>user agent</a> can associate certain
```
```suggestion
           </p><p>However, the <a>user agent</a> can associate certain
```

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          <code>*-primary</code> and <code>*-secondary</code> values at will.
+          One primary orientation will always be determined by the natural
+          orientation of the device and this will then determine the secondary
+          value of its related orientation. For example a phone held in its
+          natural orientation would have a current orientation of
+          <a>portrait-primary</a> and its <a>portrait-secondary</a> orientation
+          would be its position when rotated 180°. The decision on setting the
+          other primary and secondary orientation is up to the <a>user
+          agent</a>. For example, it can be based on the device preferred
+          angles, the user's preferred orientations or the current orientation
```
```suggestion
          angles, or the user's preferred orientations, or the current orientation
```

index.html
marcoscaceres: Avoid the word "phone"... maybe just say "device".

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          <code>270</code> or <a>portrait-primary</a> at <a>angle</a>
+          <code>90</code> or <code>270</code>. The secondary orientation is set
+          at <a>angle</a> <code>90</code> or <code>270</code> depending on what
+          was set for the related primary orientation value. In the case of the
+          above phone example, the <a>user agent</a> may determine that the
+          phone’s <a>landscape-primary</a> orientation is at <a>angle</a>
+          <code>270</code> (where the phone is rotated 90° clockwise). In this
+          case its landscape-secondary orientation is <a>angle</a>
+          <code>90</code> (where the phone is rotated 90° counterclockwise).
+          Another document MAY get the opposite relationship.
```
marcoscaceres: This MAY here is somewhat confusing, specially because it feels like it's contradicting the next paragraph.

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          <a>portrait-primary</a>. When a device is turned 180° from its
+          natural orientation, its orientation will be represented by the
+          secondary value related to its natural orientation, either
+          <a>landscape-secondary</a> or <a>portrait-secondary</a>. The <a>user
+          agent</a> sets the other primary orientation, which will be either
+          <a>landscape-primary</a> at <a>angle</a> <code>90</code> or
+          <code>270</code> or <a>portrait-primary</a> at <a>angle</a>
+          <code>90</code> or <code>270</code>. The secondary orientation is set
+          at <a>angle</a> <code>90</code> or <code>270</code> depending on what
+          was set for the related primary orientation value. In the case of the
```
```suggestion
           was set for the related primary orientation value. </p> <p>In the case of the
```

index.html
```diff
@@ -297,11 +297,43 @@ <h2>
...
+          above phone example, the <a>user agent</a> may determine that the
```
marcoscaceres: When I read "above phone example" is a bit too visual, in the sense that I went looking for an actual figure...  
  
```suggestion
          aforementioned example, the <a>user agent</a> might determine that the
```

Although, we should consider adding a figure or table. It's is admittedly hard to figure out what's going to be what.

marcoscaceres: As a general rule, avoid the RFC2119 keywords (may, must, should - even in lowercase) in examples - and other non-normative things - including when you write "For example, ...".

marcoscaceres (commented):
Some suggestions for breaking this up a bit.

Johanna-hub:
I have added an explanatory table and cut the prose explanation.  Hoping this is more useful and clear.

index.html
```diff
@@ -35,6 +35,7 @@
       caniuse: "screen-orientation",
     };
     </script>
+    <link rel="stylesheet" type="text/css" href="styles.css">
```
marcoscaceres: I'd suggest just inlining style.css instead with a `<style>` element... one less file to deal with.

index.html
```diff
@@ -297,11 +298,95 @@ <h2>
...
+          have a current orientation of <a>portrait-primary</a> and its
+          <a>portrait-secondary</a> orientation would be its position when
+          rotated 180°.
+        </p>
+        <p>
+          The <a>user agent</a> can associate the other <code>*-primary</code>
+          and <code>*-secondary</code> values at will. For example, it can be
+          based on the device preferred angles, the user's preferred
+          orientations or the current orientation when the application starts.
+        </p>
```
marcoscaceres, line 326: In a new paragraph, it would be nice here to introduce the table... "The following table..." Alternatively, you could talk about the table in the first paragraph that you added above.

index.html
```diff
@@ -297,11 +298,95 @@ <h2>
...
+              </th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>
+                Portrait
+              </td>
+              <td>
+                Portrait-primary<br>
```
marcoscaceres: nit: link "portrait-primary" etc... 

Using a `<br>` seems a little odd... we probably don't need to fix this now, but we should think about maybe a better way of arranging the cells. It doesn't matter too much tho.

index.html
```diff
@@ -297,11 +298,95 @@ <h2>
...
+                Portrait-secondary<br>
+                Set at the angle not used for portrait-primary
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          A <a>user agent</a> MUST keep the <a>current orientation type</a> and
+          the <a>current orientation angle</a> relation consistent for any
+          given <a>document</a>.
```
marcoscaceres: May the table could play a role here in enforcing this consistency?

index.html
```diff
@@ -297,11 +298,95 @@ <h2>
...
+          <a>portrait-secondary</a> orientation would be its position when
+          rotated 180°.
+        </p>
+        <p>
+          The <a>user agent</a> can associate the other <code>*-primary</code>
+          and <code>*-secondary</code> values at will. For example, it can be
+          based on the device preferred angles, the user's preferred
+          orientations or the current orientation when the application starts.
+        </p>
+        <table>
```
marcoscaceres, line 333: Might be nice to give the table a `<caption><dfn>the xxxxx table</dfn></caption>` with a proper name, allowing it to be referenced from within the document. Then you can do, "In `<a>`the xxxxx table</a>" when needed (and also potentially allowing other specs to externally reference the table).

marcoscaceres (changes requested):
Couple of little things, but this is looking really good! Much more clear with the table.

index.html
marcoscaceres: If you name the table via the caption, as suggested below, then you can just reference the table by name instead of the "in the following table".

marcoscaceres (approved):
The new "screen orientation values table" and intro paragraph are really good. Nice and clear!

PR #132, opened 2019-01-10
Editorial: Introduction Section

Johanna-hub:
For comments/review 🙂

I've had a first go at drafting the introductory section as discussed

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/132.html" title="Last updated on Jan 21, 2019, 9:43 PM UTC (a147af4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/132/c904bd4...Johanna-hub:a147af4.html" title="Last updated on Jan 21, 2019, 9:43 PM UTC (a147af4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,11 +76,16 @@
         Introduction
       </h2>
       <p>
-        The Screen Orientation API provides the current screen orientation and
-        any orientation changes so all web applications can display correctly
-        and update where needed. It enables web applications that provide the
-        most optimal user experience in one particular screen orientation to
-        lock to that orientation.
+        For web applications, the Screen Orientation API exposes the type and
+        angle of a device's current screen orientation, and can provide
+        notification if the device's orientation changes. This allows web
+        applications to programmatically adapt the user experience for many
+        possible screen orientations (in concert with CSS). The API also
+        provides web applications with the ability to lock the screen to a
+        particular orientation, which is needed for an optimal user experience
+        in some types of applications. This is useful in applications like
+        computer games, where the user can physically rotate the device as a
+        means of user input but the screen orientation itself mustn't change.
       </p>
       <p>
         This specification describes how the API:
--- a/index.html
+++ b/index.html
@@ -94,14 +94,14 @@
         <li>Provides the current screen orientation type and angle of a device
         as specified by the <a>updating orientation algorithm</a>.
         </li>
-        <li>Updates the screen orientation when the viewport angle changes
-        as specified by the <a>screen orientation change algorithm</a>.
+        <li>Updates the screen orientation when the viewport angle changes as
+        specified by the <a>screen orientation change algorithm</a>.
         </li>
-        <li>Locks the screen to a specific orientation as specified by the <a>locking
-        orientation algorithm</a>.
+        <li>Locks the screen to a specific orientation as specified by the <a>
+          locking orientation algorithm</a>.
         </li>
-        <li>Unlocks the locked screen to the default orientation as specified by
-        the <a data-link-for="ScreenOrientation">unlock()</a> method.
+        <li>Unlocks the locked screen to the default orientation as specified
+        by the <a data-link-for="ScreenOrientation">unlock()</a> method.
         </li>
       </ul>
       <section>
--- a/index.html
+++ b/index.html
@@ -72,28 +72,47 @@
       </p>
     </section>
     <section class="informative">
-      <h2>Introduction</h2>
+      <h2>
+        Introduction
+      </h2>
       <p>
-          The Screen Orientation API allows web applications that would be negatively impacted by users changing the orientation of their device to block any changes.  It also provides the current screen orientation and any orientation changes so all web applications can display correctly.
+        The Screen Orientation API allows web applications that would be
+        negatively impacted by users changing the orientation of their device
+        to block any changes. It also provides the current screen orientation
+        and any orientation changes so all web applications can display
+        correctly.
       </p>
       <p>
-          This specification describes how the API:
-        </p>
-          <ul>
-          <li>provides the current screen orientation type and angle of a device as defined in the <a>updating orientation algorithm</a>.</li>
-          <li>updates the screen orientation when the viewport angle changes according to the <a>screen orientation change algorithm</a>.</li>
-          <li>locks the screen to a specific orientation following the <a>locking orientation algorithm</a>.</li>
-          <li>unlocks the locked screen to the default orientation as described in the <a data-link-for="ScreenOrientation">unlock()</a> method.</li>
-        </ul>
-   
+        This specification describes how the API:
+      </p>
+      <ul>
+        <li>provides the current screen orientation type and angle of a device
+        as defined in the <a>updating orientation algorithm</a>.
+        </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -71,6 +71,51 @@
         This document is a work in progress.
       </p>
     </section>
+    <section class="informative">
+      <h2>
+        Introduction
+      </h2>
+      <p>
+        The Screen Orientation API allows web applications that would be
```
marcoscaceres: It might be nice to to move the last sentence first (what the API does) and then describe what affordances the API provides. Also, it might be better to frame this in the positive sense - so, not so much "negatively impacted", but more that that some applications provide a "more optional user experience" when used in particular orientations (e.g., some games in landscape).

index.html
```diff
@@ -71,6 +71,51 @@
...
+        as defined in the <a>updating orientation algorithm</a>.
+        </li>
+        <li>updates the screen orientation when the viewport angle changes
+        according to the <a>screen orientation change algorithm</a>.
+        </li>
+        <li>locks the screen to a specific orientation following the <a>locking
+        orientation algorithm</a>.
+        </li>
+        <li>unlocks the locked screen to the default orientation as described
+        in the <a data-link-for="ScreenOrientation">unlock()</a> method.
```
marcoscaceres: Don't we have an unlock algorithm? Or are the steps to unlock explicitly part of the .unlock()?

index.html
```diff
@@ -71,6 +71,51 @@
...
+        </li>
+      </ul>
+      <section>
+        <h2>
+          Goals
+        </h2>
+        <ul>
+          <li>Allow a web application to have control over screen orientation
```
marcoscaceres, line 112: We probably want to say something about "without being intrusive to end users". Like, a developer can't randomly change screen orientation whenever they want... only when the security conditions are met. That's the goal :)

index.html
```diff
@@ -71,6 +71,51 @@
...
+          where needed.
+          </li>
+          <li>Support a better user experience by giving information on screen
+          orientation changes so web applications can update correspondingly.
+          </li>
+          <li>Provide a simple way for users to understand how to hold their
```
marcoscaceres: I think this goal needs to be phrased with a developer focus. As it reads, this would be a goal if the API gave hints to the user that the orientation was about to change - but we don't do that. We leave it to the OS. 

However, actually letting the developer know when the orientation has changed, and providing the developer with both the updated orientation and angle, so that they can make necessary adjustments to their web application is definitely a goal.

marcoscaceres (changes requested):
Looking good. You've captured the intent of the API quite nicely. Some suggestions for improvements.

index.html
Johanna-hub: No there is no unlock algorithm.  The unlock() method basically says to run the steps to lock and set to default orientation.

index.html
```diff
@@ -76,11 +76,11 @@ <h2>
         Introduction
       </h2>
       <p>
-        The Screen Orientation API allows web applications that would be
-        negatively impacted by users changing the orientation of their device
-        to block any changes. It also provides the current screen orientation
-        and any orientation changes so all web applications can display
-        correctly.
+        The Screen Orientation API provides the current screen orientation and
```
marcoscaceres: The thing we need to be careful of is making claims about being "correct". I'd also like to make sure the intro covers: angle, type, event - and why... and also that this is not a substitute for CSS (the locking aspect is the most important with that regards... CSS can't lock, but CSS does the actual content adaptation). Most of that is in what you wrote, but how about (feel free to change!): 

>  For web applications, the Screen Orientation API exposes the type and angle of a device's current screen orientation, as well as the ability to be notified if the device's orientation changes. This allows web applications to programmatically adapt the user experience for many possible screen orientations (in concert with CSS). The API also affords web applications with the ability to lock the screen to a particular orientation, which can provide for an optimal user experience in some types of applications. This is useful in applications like computer games, where the user can physically rotate the device as a means of user input but the screen orientation itself mustn't  change.

marcoscaceres (changes requested):
Suggestion to work off...

index.html
marcoscaceres: I wonder if we should abstract out the unlocking steps from the method? Do any algorithms  unlock by calling into .unlock()?

Johanna-hub: I agree it probably makes sense to have unlock() as an algorithm.  I'll leave it as it is for now, and when I have written it, I will change the wording/link

marcoscaceres: yeah, we can do it as a new PR.

index.html
Johanna-hub: This is now edited, please review the intro as a whole, think it's all ready

index.html
```diff
@@ -71,6 +71,54 @@
...
+        particular orientation, which is needed for an optimal user experience
+        in some types of applications. This is useful in applications like
+        computer games, where the user can physically rotate the device as a
+        means of user input but the screen orientation itself mustn't change.
+      </p>
+      <p>
+        This specification describes how the API:
+      </p>
+      <ul>
+        <li>provides the current screen orientation type and angle of a device
```
```suggestion
        <li>Provides the current screen orientation type and angle of a device,
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        as defined in the <a>updating orientation algorithm</a>.
+        </li>
+        <li>updates the screen orientation when the viewport angle changes
```
```suggestion
        <li>Updates the screen orientation when the viewport angle changes,
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        in some types of applications. This is useful in applications like
+        computer games, where the user can physically rotate the device as a
+        means of user input but the screen orientation itself mustn't change.
+      </p>
+      <p>
+        This specification describes how the API:
+      </p>
+      <ul>
+        <li>provides the current screen orientation type and angle of a device
+        as defined in the <a>updating orientation algorithm</a>.
```
```suggestion
        as specified by the <a>updating orientation algorithm</a>.
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        </li>
+        <li>updates the screen orientation when the viewport angle changes
+        according to the <a>screen orientation change algorithm</a>.
```
```suggestion
        as specified by the <a>screen orientation change algorithm</a>.
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        </li>
+        <li>locks the screen to a specific orientation following the <a>locking
```
```suggestion
        <li>locks the screen to a specific orientation, as specified by the <a>locking
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        orientation algorithm</a>.
+        </li>
+        <li>unlocks the locked screen to the default orientation as described
```
```suggestion
        <li>Unlocks the locked screen to the default orientation,
```

index.html
```diff
@@ -71,6 +71,54 @@
...
+        in the <a data-link-for="ScreenOrientation">unlock()</a> method.
```
```suggestion
        as as specified by the <a data-link-for="ScreenOrientation">unlock()</a> method.
```

marcoscaceres (approved):
Small nits. Will probably need a tidy after applying.

PR #133, opened 2019-01-15
Editorial: Example 1

Johanna-hub:
For review

I have changed this to be more specific to a phone because I think this makes the example clearer and is likely to be how people are testing out the example.  I have given it the specific aim of locking to landscape because I think this is also more useful.

(The full examples are available with a link for testing on phones at https://github.com/Johanna-hub/screen-orientation-examples)

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/133.html" title="Last updated on Jan 17, 2019, 1:24 PM UTC (4b2c2d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/133/c904bd4...Johanna-hub:4b2c2d4.html" title="Last updated on Jan 17, 2019, 1:24 PM UTC (4b2c2d4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -541,9 +541,9 @@
       <pre class='example html'>
 &lt;script&gt;
 async function fullScreenCheck() {
-  if(document.fullscreenElement === null) {
-      await document.documentElement.requestFullscreen();
-  } 
+  if (document.fullscreenElement === null) {
+    await document.body.requestFullscreen();
+  }
 }
 
 async function unlock() {
@@ -563,7 +563,7 @@ async function rotate() {
 
 function show() {
   const { type, angle } = screen.orientation;
-  console.log(`Orientation type is ${type} + angle is ${angle}`);
+  console.log(`Orientation type is ${type} & angle is ${angle}`);
 }
 
 screen.orientation.addEventListener("change", show);
--- a/index.html
+++ b/index.html
@@ -570,7 +570,7 @@ screen.orientation.addEventListener("change", show);
 window.addEventListener("load", show);
 &lt;/script&gt;
 
-&lt;button onclick="rotate()"&gt;
+&lt;button onclick="rotate() id="rotate"&gt;
   Rotate
 &lt;/button&gt;
 &lt;button onclick='unlock()'&gt;
--- a/index.html
+++ b/index.html
@@ -529,10 +529,10 @@
         Examples
       </h2>
       <p>
-        In this example for a mobile device, clicking the Rotate button
-        makes a request to go fullscreen and then lock to the opposite
-        orientation. If the device is not in it's default orientation, pressing the unlock button
-        reverts the screen orientation back to default.
+        In this example for a mobile device, clicking the Rotate button makes a
+        request to go fullscreen and then lock to the opposite orientation. If
+        the device is not in it's default orientation, pressing the unlock
+        button reverts the screen orientation back to default.
       </p>
       <p>
         The log shows the change in orientation type and angle.
--- a/index.html
+++ b/index.html
@@ -529,11 +529,15 @@
         Examples
       </h2>
       <p>
-        In this example for a mobile phone,  clicking the landscape button makes a request to go fullscreen and then lock to landscape orientation.  
-        Once it is in landscape, the unlock button will take the screen orientation back to default (portrait) or the portrait button will also change it back to portrait.
+        In this example for a mobile phone, clicking the landscape button makes
+        a request to go fullscreen and then lock to landscape orientation. Once
+        it is in landscape, the unlock button will take the screen orientation
+        back to default (portrait) or the portrait button will also change it
+        back to portrait.
       </p>
       <p>
-        The log shows the change in orientation type and angle, and whether the event type is load or change.
+        The log shows the change in orientation type and angle, and whether the
+        event type is load or change.
       </p>
       <pre class='example html'>
 &lt;script&gt;
--- a/index.html
+++ b/index.html
@@ -531,8 +531,8 @@
       <p>
         In this example for a mobile device, clicking the Rotate button makes a
         request to go fullscreen and then lock to the opposite orientation. If
-        the device is not in it's default orientation, pressing the unlock
-        button reverts the screen orientation back to default.
+        the screen is not in default orientation, pressing the unlock button
+        reverts the screen orientation back.
       </p>
       <p>
         The log shows the change in orientation type and angle.
--- a/index.html
+++ b/index.html
@@ -529,28 +529,46 @@
         Examples
       </h2>
       <p>
-        This example shows the current screen orientation to the console every
-        time the screen orientation state changes.
+        In this example for a mobile phone,  clicking the landscape button makes a request to go fullscreen and then lock to landscape orientation.  
+        Once it is in landscape, the unlock button will take the screen orientation back to default (portrait) or the portrait button will also change it back to portrait.
+      </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -529,28 +529,50 @@ <h2>
         Examples
       </h2>
       <p>
-        This example shows the current screen orientation to the console every
-        time the screen orientation state changes.
+        In this example for a mobile phone, clicking the landscape button makes
```
```suggestion
        In this example for a mobile device, clicking the landscape button makes
```

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
+        a request to go fullscreen and then lock to landscape orientation. Once
+        it is in landscape, the unlock button will take the screen orientation
```
```suggestion
        it is in landscape, pressing the unlock button reverts the screen orientation
```

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
+        back to default (portrait) or the portrait button will also change it
```
```suggestion
        back to default (portrait). Pressing the button labelled "portrait" will also change it
```

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
+        back to portrait.
```
```suggestion
        back to <a>portrait</a>.
```

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
+      </p>
+      <p>
+        The log shows the change in orientation type and angle, and whether the
+        event type is load or change.
```
```suggestion
        event type is "load" or "change".
```

marcoscaceres: Or perhaps, ", and the type of event."

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
       </p>
       <pre class='example html'>
 &lt;script&gt;
-  var show = function() {
-     console.log("Orientation type is " + screen.orientation.type);
-     console.log("Orientation angle is " + screen.orientation.angle);
-  }
+async function lockLandscape(){
+  await document.documentElement.requestFullscreen()
```
```suggestion
  await document.documentElement.requestFullscreen();
```

marcoscaceres: I think we might actually want to do the `checkIfFullScreen()` function, otherwise, it looks like you always have to call requestFullScreen() before using the API. 

For `checkIfFullScreen()`, you will need to figure out how to check if fullscreen is enabled (MDN should probably be able to tell you). Basically, if fullscreen is already enabled, then just return, otherwise, call requestFullscreen(); and return the resulting promise.

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
+  await document.documentElement.requestFullscreen() 
+  await screen.orientation.lock('landscape');
+}
+
+async function unlock() {
+  await screen.orientation.unlock();
+}
+
+async function portrait(){
+  await screen.orientation.lock('portrait');
```
marcoscaceres: Nit: here you need the full screen check... otherwise, this will fail if the button is just pressed.

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
-  window.onload = show;
+const { type, angle } = screen.orientation
+function show(event) {
+ console.log("Orientation type is " + screen.orientation.type);
+ console.log("Orientation angle is " + screen.orientation.angle);
+ console.log("Event Type is " + event.type)
+}
+
+screen.orientation.addEventListener("change", show);
+window.onload = show;
```
marcoscaceres: Just to be consistent with previous line: 

```suggestion
window.addEventListener("load", show);
```

index.html
```diff
@@ -529,28 +529,50 @@ <h2>
...
 &lt;/button&gt;
-&lt;button onclick="screen.orientation.lock('portrait')"&gt;
-  Lock to portrait
+&lt;button onclick='unlock()'&gt;
+  Unlock
 &lt;/button&gt;
-&lt;button onclick="screen.orientation.lock('landscape')"&gt;
-  Lock to landscape
+&lt;button onclick="portrait()"&gt;
+  Portrait
```
marcoscaceres: What might be cute is to "rotate" instead... you could check the orientation, then switch from one to the other.

marcoscaceres (commented):
Rapid fire suggestions :)

index.html
Johanna-hub: I have added these in, just wanted to explain the reason I didn't put it in initially.  On my phone once you have locked to landscape you can only stay in full screen.  If you try to exit full screen it just takes you back to portrait.  So, for this specific example, where we are locking to landscape and then unlocking, we don't need the check.  Other phones might be different of course, so probably better to put it in! 😊

Johanna-hub: Sure! Let's chat about this tomorow.

index.html
Johanna-hub: Sure! Let's chat about this tomorrow!

index.html
```diff
@@ -529,28 +529,58 @@ <h2>
...
+}
+
+async function portrait(){
+  await fullScreenCheck();
+  await screen.orientation.lock('portrait');
+}
 
-  screen.orientation.addEventListener("change", show);
-  window.onload = show;
+const { type, angle } = screen.orientation
```
marcoscaceres: This could be inside the show() function, then you can just:

```JS
const { type, angle } = screen.orientation
console.log("Orientation type is " + type);
console.log("Orientation angle is " + angle);
```

marcoscaceres: Might be less redundant to do:

```JS
const { type, angle } = screen.orientation
console.log(`Orientation type Is "${type}", angle is ${angle}.`);
```

index.html
```diff
@@ -529,28 +529,58 @@ <h2>
...
+function show(event) {
+ console.log("Orientation type is " + screen.orientation.type);
+ console.log("Orientation angle is " + screen.orientation.angle);
+ console.log("Event Type is " + event.type)
```
marcoscaceres: We should drop showing the event.type, I think. It doesn't really add much.

marcoscaceres (commented):
Couple of suggestions...

index.html
```diff
@@ -529,28 +529,58 @@ <h2>
...
+        event.
       </p>
       <pre class='example html'>
 &lt;script&gt;
-  var show = function() {
-     console.log("Orientation type is " + screen.orientation.type);
-     console.log("Orientation angle is " + screen.orientation.angle);
-  }
+async function fullScreenCheck() {
+  if(document.fullscreenElement === null) {
```
```suggestion
  if (document.fullscreenElement === null) {
```

marcoscaceres: Make sure these code blocks are prettied ✨

index.html
```diff
@@ -529,28 +529,52 @@ <h2>
...
+    : "portrait";
+  await screen.orientation.lock(newOrientation);
+  rotate.textContent = `Rotate to ${newOrientation}`;
+}
 
-  screen.orientation.addEventListener("change", show);
-  window.onload = show;
+function show() {
+  const { type, angle } = screen.orientation;
+  console.log(`Orientation type is ${type} + angle is ${angle}`);
```
```suggestion
  console.log(`Orientation type is ${type} & angle is ${angle}`);
```

marcoscaceres (approved):
Couple of nits, but these are looking pretty solid.

Johanna-hub:
@marcoscaceres I realised that the explanatory text hasn't been changed since we changed the button to just rotate and unlock!  Can you just check the new wording before I merge 😊

marcoscaceres (approved):
new text is good. Thanks for updating it.

PR #135, opened 2019-01-16
Editorial: Example 3

Johanna-hub:
For review

- Main change from the original example, if screen orientation is not supported then `screen.orientation.type.startsWith` and `screen.orientation.add/remove.Event.Listener` should also not be used. 

- I changed the event listener to `window.add/removeEventListener("orientationchange")`.

- I used `window.matchMedia("(orientation: portrait)").matches` to check the orientation before displaying the final alert. However I have tested on iOS and android with differing results!  Available to test here https://johanna-hub.github.io/example-3/

- Safari and Chrome on iOS the example works completely fine.
- Chrome and Firefox on Android, the final completed alert only comes up once you have rotated back to portrait (same thing happens in devtools responsive mode on Chrome)

I think the matchMedia querylist needs a listener but then it gets too complicated!  So might be better to do `if(screen.width>screen.height)` just for the purposes of demonstration of the concept.

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/135.html" title="Last updated on Jan 22, 2019, 11:22 AM UTC (f3f1b5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/135/c904bd4...Johanna-hub:f3f1b5e.html" title="Last updated on Jan 22, 2019, 11:22 AM UTC (f3f1b5e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -572,10 +572,10 @@
 
 </pre>
       <p>
-        In this example for a mobile device, if the Screen
-        Orientation API is not supported, or the document is not fullscreen and
-        the screen orientation lock rejects, the user is alerted to rotate
-        their screen manually to landscape.
+        In this example for a mobile device, if the Screen Orientation API is
+        not supported, or the document is not fullscreen and the screen
+        orientation lock rejects, the user is alerted to rotate their screen
+        manually to landscape.
       </p>
       <pre class='example html'>
 &lt;script&gt;
--- a/index.html
+++ b/index.html
@@ -572,9 +572,10 @@
 
 </pre>
       <p>
-        In this example for a mobile device held in portrait, if the Screen Orientation API is not supported, or the
-        document is not fullscreen and the screen orientation lock rejects, the
-        user is alerted to rotate their screen manually to landscape.
+        In this example for a mobile device held in portrait, if the Screen
+        Orientation API is not supported, or the document is not fullscreen and
+        the screen orientation lock rejects, the user is alerted to rotate
+        their screen manually to landscape.
       </p>
       <pre class='example html'>
 &lt;script&gt;
--- a/index.html
+++ b/index.html
@@ -579,22 +579,31 @@
       <pre class='example html'>
 &lt;script&gt;
 function manualRotation() {
-  alert("Rotation to landscape complete");
+  alert("Thank you for rotating!");
 }
 
-function orientationChangeHandler() {
-  if (window.matchMedia("(orientation: portrait)").matches) {
-    return;
+function orientationChangeListener() {
+  const inPortrait = window.matchMedia("(orientation: portrait)");
+  function detectOrientation(e) {
+    if (e.matches) {
+      return;
+    } else {
+      window.removeEventListener(
+        "orientationchange",
+        orientationChangeListener
+      );
+      inPortrait.removeListener(detectOrientation);
+      manualRotation();
+    }
   }
-  window.removeEventListener("orientationchange", orientationChangeHandler);
-  manualRotation();
+  inPortrait.addListener(detectOrientation);
 }
 
 async function start() {
   try {
     await screen.orientation.lock("landscape");
   } catch {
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):
Looks good!

Johanna-hub:
@marcoscaceres I've fixed this by adding the listener to matchMedia and it's working across all browsers on my phone.  Can you check that it hasn't become too long now.  Your suggestion on Slack didn't work unfortunately.

index.html
```diff
@@ -572,31 +572,45 @@ <h2>
...
-          }
-          screen.orientation.removeEventListener('change', orientationChangeHandler);
-          startInternal();
-        }
+function orientationChangeListener() {
+  const inPortrait = window.matchMedia("(orientation: portrait)");
+  function detectOrientation(e) {
+    if (e.matches) {
+      return;
+    } else {
```
marcoscaceres: nit: ` else {`  is not needed here, because you are returning immediately.

Johanna-hub:
@marcoscaceres Can you just double check my changes from Slack before I merge, thanks!

index.html
```diff
@@ -572,31 +572,44 @@ <h2>
...
-          if (!screen.orientation.type.startsWith('landscape')) {
-            return;
-          }
-          screen.orientation.removeEventListener('change', orientationChangeHandler);
-          startInternal();
+async function start() {
+  try {
+    await screen.orientation.lock("landscape");
+  } catch {
+    const inPortrait = window.matchMedia("(orientation: portrait)");
```
marcoscaceres: This seems to assume that the user is initially holding the device in portrait. However, they might already be in landscape mode.

index.html
```diff
@@ -572,31 +572,44 @@ <h2>
...
+    function orientationChangeListener() {
+      function detectOrientation(e) {
+        if (e.matches) {
```
marcoscaceres: Similarly here, it could be that other orientations will match, not just "portrait"... so this should actually be checking for "matches === landscape".

index.html
```diff
@@ -572,31 +572,44 @@ <h2>
...
+          "orientationchange",
+          orientationChangeListener
+        );
+        inPortrait.removeListener(detectOrientation);
+        manualRotation();
+      }
+      inPortrait.addListener(detectOrientation);
+    }
+    window.addEventListener("orientationchange", orientationChangeListener);
+    alert("To start, please rotate your screen to landscape");
```
marcoscaceres: Squishing all of the above, I get to:

```JS
function start() {
  /* Start application when in correct orientation */
}
async function rotate() {
  try {
    await screen.orientation.lock("landscape");
    start();
  } catch (err) {
    console.error(err);
  }
  // Not supported or not working, so...
  const matchLandscape = matchMedia("(orientation: landscape)");
  if (matchLandscape.matches) return start(); // Already landscape!
  // Let's get the user to rotate manually...
  addEventListener("orientationchange", function listener() {
    matchPortrait.addListener(function mediaChange(e) {
      if (!e.matches) return; // Nope, not landscape!
      removeEventListener("orientationchange", listener);
      matchPortrait.removeListener(mediaChange);
      start();
    });
  });
  alert("To start, please rotate your screen to landscape.");
}
```

index.html
```diff
@@ -572,31 +572,44 @@ <h2>
...
       </p>
-      <pre class='example js'>
+      <pre class='example html'>
 &lt;script&gt;
-  var start = function() {
-    screen.orientation.lock('landscape-primary').then(
-      startInternal,
-      function() {
-        alert('To start, rotate your screen to landscape.');
+function manualRotation() {
```
marcoscaceres: I think we should just inline this (see below).

marcoscaceres (changes requested):
Some changes for us to discuss...

marcoscaceres (approved):
Thanks for trying this and fixing.

PR #137, opened 2019-01-23
Editorial: Examples section edits

Johanna-hub:
In testing Example 1 with the discussed revisions, I came across some issues so a lot of it has been rewritten, so please give feedback!

- we changed document.documentElement.requestFullScreen() to document.body.requestFullScreen() for reasons of length.  This actually makes the whole screen go black so I've changed it back to solve that problem
- The way I had written the example before, the button text only changed when it was pressed.  I have now changed it so that it also changes on user rotation (otherwise user could rotate manually to portrait and it would still say lock to portrait).  It also displays the opposite orientation on load.

Other example edits have been completed, sections/titles, edits and moving it up to Introduction section.

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/137.html" title="Last updated on Jan 29, 2019, 11:34 AM UTC (e833f95)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/137/98b4058...Johanna-hub:e833f95.html" title="Last updated on Jan 29, 2019, 11:34 AM UTC (e833f95)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -106,7 +106,6 @@
       </ul>
       <section>
         <h2>
-
           Goals
         </h2>
         <ul>
@@ -124,17 +123,20 @@
           Examples
         </h2>
         <section>
-          <h2>Locking to a specific orientation and unlocking</h2>
-        <p>
-          In this example, clicking the "Lock" button makes a
-          request to go into fullscreen and then lock to the opposite orientation. If
-          the screen is not in default orientation, pressing the "Unlock" button
-          unlocks to its default orientation.  
-        </p>
-        <p>
-          The developer console logs the change in orientation type and angle.
-        </p>
-        <pre class='example html'>
+          <h2>
+            Locking to a specific orientation and unlocking
+          </h2>
+          <p>
+            In this example, clicking the "Lock" button makes a request to go
+            into fullscreen and then lock to the opposite orientation. If the
+            screen is not in default orientation, pressing the "Unlock" button
+            unlocks to its default orientation.
+          </p>
+          <p>
+            The developer console logs the change in orientation type and
+            angle.
+          </p>
+          <pre class='example html'>
   &lt;script&gt;
   function fullScreenCheck() {
     if (document.fullscreenElement) return;
@@ -185,14 +187,16 @@
     Unlock
   &lt;/button&gt;
   </pre>
-</section>
-<section>
-  <h2>Locking the screen before calling a function</h2>
-        <p>
-          This example waits to go into fullscreen, then locks to landscape before
-          calling <code>ready()</code>.
-        </p>
-        <pre class='example html'>
+        </section>
+        <section>
+          <h2>
+            Locking the screen before calling a function
+          </h2>
+          <p>
+            This example waits to go into fullscreen, then locks to landscape
+            before calling <code>ready()</code>.
+          </p>
+          <pre class='example html'>
   &lt;script&gt;
   function ready() {
     const { type } = screen.orientation;
@@ -209,16 +213,18 @@
     Start
   &lt;/button&gt;
   </pre>
-</section>
-<section>
-  <h2>Alerting the user if the Screen Orientation API is not supported</h2>
-        <p>
-          In this example, if the Screen Orientation API is
-          not supported, or the document is not fullscreen and the screen
-          orientation lock rejects, the user is alerted to rotate their screen
-          manually to landscape.
-        </p>
-        <pre class='example html'>
+        </section>
+        <section>
+          <h2>
+            Alerting the user if the Screen Orientation API is not supported
+          </h2>
+          <p>
+            In this example, if the Screen Orientation API is not supported, or
+            the document is not fullscreen and the screen orientation lock
+            rejects, the user is alerted to rotate their screen manually to
+            landscape.
+          </p>
+          <pre class='example html'>
   &lt;script&gt;
   function start() {
     /* Start application when in correct orientation */
@@ -247,7 +253,7 @@
     Start
   &lt;/button&gt;
   </pre>
-</section>
+        </section>
       </section>
     </section>
     <section data-dfn-for='Screen'>
--- a/index.html
+++ b/index.html
@@ -106,6 +106,7 @@
       </ul>
       <section>
         <h2>
+
           Goals
         </h2>
         <ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+  const lockButton = document.getElementById("lockBtn");
+
+  function btnOrientation() {
+    const btnOrientation = oppOrientation();
+    lockButton.textContent = `Lock to ${btnOrientation}`;
+  }
+  
+  function oppOrientation() {
+    const { type } = screen.orientation;
+    if (type.startsWith("portrait")) {
```
marcoscaceres: Nit: maybe just... 
```JS
return type.startsWith("portrait") ? "landscape" : "portrait";
```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+    const { type, angle } = screen.orientation;
+    console.log(`Orientation type is ${type} & angle is ${angle}.`);
+  }
+  
+  screen.orientation.addEventListener("change", show);
+  screen.orientation.addEventListener("change", btnOrientation);
+  window.addEventListener("load", show);
+  &lt;/script&gt;
+  
+  &lt;button onclick="rotate() id="lockBtn"&gt;
```
marcoscaceres: If we pass "this", then we don't need to look up the button. 
```suggestion
  &lt;button onclick="rotate(this)"&gt;
```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+  
+  function oppOrientation() {
+    const { type } = screen.orientation;
+    if (type.startsWith("portrait")) {
+      return "landscape";
+    }
+    return "portrait";
+  }
+
+  async function rotate() {
```
marcoscaceres: See below... we can just pass the button that was clicked.  
```suggestion
  async function rotate(rotateBtn) {
```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+    const rotateBtn = document.getElementById("rotateBtn");
```
marcoscaceres: See below... 
```suggestion

```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+    await screen.orientation.lock(newOrientation);
+    btnOrientation();
+  }
+  
+  function show() {
+    const { type, angle } = screen.orientation;
+    console.log(`Orientation type is ${type} & angle is ${angle}.`);
+  }
+  
+  screen.orientation.addEventListener("change", show);
```
marcoscaceres: combine these two into a singe
```JS
addEventListener("change", ()=>{
   show();
   btnOrientation(); 
}) 
```
That also assures the order means that the browser doesn't need to fire two separate events.

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+          <pre class='example html'>
+  &lt;script&gt;
+  function fullScreenCheck() {
+    if (document.fullscreenElement) return;
+    return document.documentElement.requestFullscreen();
+  }
+  
+  const lockButton = document.getElementById("lockBtn");
+
+  function btnOrientation() {
```
marcoscaceres: Nit - make sure every functions name is a verb (and clearly indicates what action it does), and not a noun. Also, it's always best to pass things into a function, and not rely on the global scope. 
  
```suggestion
  function updateDetails(lockButton) {
```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+    const btnOrientation = oppOrientation();
+    lockButton.textContent = `Lock to ${btnOrientation}`;
+  }
+  
+  function oppOrientation() {
```
marcoscaceres: As above with the name... maybe `getOppositeOrientation`. Generally, try to avoid abbreviations in names (applies globally as a comment).

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+            angle.
+          </p>
+          <pre class='example html'>
+  &lt;script&gt;
+  function fullScreenCheck() {
+    if (document.fullscreenElement) return;
+    return document.documentElement.requestFullscreen();
+  }
+  
+  const lockButton = document.getElementById("lockBtn");
```
marcoscaceres: This will be "racy" because lockButton doesn't actually exist yet (the HTML parser has not yet parsed button underneath). So this will be null.  

I'd suggest deleting it, and query the doc only when you need this. 

```suggestion

```

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+
+  function btnOrientation() {
+    const btnOrientation = oppOrientation();
+    lockButton.textContent = `Lock to ${btnOrientation}`;
```
marcoscaceres: I think this is not correct. As we discussed yesterday, `.unlock()` doesn't mean it will revert the orientation. Thus, the `lockButton`'s text should always just be "Unlock". However, the "Rotate" button should change its text...

marcoscaceres (changes requested):
Looking good, but example 1 needs a bit of a rework...

Johanna-hub:
I have incorporated the changes and tested again on my phone.  

I had to leave an id on the button because we need to change the text on the button on load and on manual rotation, so can't just rely on the click.  I only look up the id on the associated event listeners.

index.html
```diff
@@ -118,6 +118,143 @@ <h2>
...
+    updateDetails(document.getElementById("button"));
+  });
+  
+  window.addEventListener("load", () => {
+    show();
+    updateDetails(document.getElementById("button"));
+  });
+  &lt;/script&gt;
+  
+  &lt;button onclick="rotate() id="button"&gt;
```
marcoscaceres: Note that this was still buggy... missing closing quote and we should pass in this.  

```suggestion
  &lt;button onclick="rotate(this)" id="button"&gt;
```

marcoscaceres (approved):
I think there might still be a tiny bug in the button. See suggestion. The other examples seem fine.

I think we should let Example 1 "cook" 🥘 for a bit... I'm still seeing places were we could make a few tweaks. I'd like to come back and revise it a bit later in the internship. However, we got more fun/pressing things to focus on :)

Johanna-hub:
> I think we should let Example 1 "cook" 🥘 for a bit... I'm still seeing places were we could make a few tweaks. I'd like to come back and revise it a bit later in the internship. However, we got more fun/pressing things to focus on :)

Is it ok to merge now and work on Example 1 later or should we wait?

marcoscaceres:
Totally fine to merge!

PR #147, opened 2019-02-04
Resolving the promise after the change event is fired

### LINKED ISSUE #120: Send events before promise resolve
An annoyance with the orientation lock API, is that the promise resolves before the event is sent.

This makes it hard to default handlers that you don't want to call in certain cases, because if you remove the handler then there is no good place to add it back in, because if you do after the promise gets resolves, it might get called as the event is still pending :-( 

This also means that it makes sense for the unlock to return a promise.

We ran into this in https://github.com/intel/generic-sensor-demos/issues/72

Johanna-hub:
Closes #120

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [x] Modified Web platform tests https://github.com/web-platform-tests/wpt/pull/15267

Implementation commitment:

 * [ ] WebKit 
 * [x] [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=930237)
 * [x] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1525582)

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/147.html" title="Last updated on Feb 19, 2019, 10:38 AM UTC (19e5f68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/147/e41f97e...Johanna-hub:19e5f68.html" title="Last updated on Feb 19, 2019, 10:38 AM UTC (19e5f68)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1005,15 +1005,6 @@
               <li>
                 <a>Update the orientation information</a> of <var>doc</var>.
               </li>
-              <li>If <var>doc</var>'s <a>pending promise</a> is not
-              <code>null</code>:
-                <ol>
-                  <li>Resolve <var>doc</var>'s <a>pending promise</a> with
-                  <code>undefined</code>.
-                  </li>
-                  <li>Set <var>doc</var>'s <a>pending promise</a> to
-                  <code>null</code> .
-                  </li>
                 </ol>
               </li>
               <li>If the orientation change was triggered by a user gesture
@@ -1026,6 +1017,15 @@
                 <a>Fire an event</a> named <code>change</code> at
                 <var>doc</var>'s <a>screen.orientation</a> object.
               </li>
+              <li>If <var>doc</var>'s <a>pending promise</a> is not
+                <code>null</code>:
+                  <ol>
+                    <li>Resolve <var>doc</var>'s <a>pending promise</a> with
+                    <code>undefined</code>.
+                    </li>
+                    <li>Set <var>doc</var>'s <a>pending promise</a> to
+                    <code>null</code> .
+                    </li>
             </ol>
           </li>
         </ol>
@@ -1048,6 +1048,16 @@
           <a>document</a>'s <a>current orientation angle</a>, run the following
           sub-steps:
             <ol>
+              <li>If the orientation change was triggered by a user gesture
+              such as the user turning the device, as opposed to a call to <a>
+                lock()</a>, the <a>task</a> MUST be annotated with
+                <code>process user orientation change</code> when running the
+                next step.
+              </li>
+              <li>
+                <a>Fire an event</a> named <code>change</code> at the
+                <a>document</a>'s <a>screen.orientation</a> object.
+              </li>
               <li>If the <a>document</a>'s <a>pending promise</a> is not <code>
                 null</code>:
                 <ol>
@@ -1058,16 +1068,6 @@
                   <code>null</code>.
                   </li>
                 </ol>
-              </li>
-              <li>If the orientation change was triggered by a user gesture
-              such as the user turning the device, as opposed to a call to <a>
-                lock()</a>, the <a>task</a> MUST be annotated with
-                <code>process user orientation change</code> when running the
-                next step.
-              </li>
-              <li>
-                <a>Fire an event</a> named <code>change</code> at the
-                <a>document</a>'s <a>screen.orientation</a> object.
               </li>
             </ol>
           </li>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Looks good. I think the gh-pages branch itself needs a tidy tho. I see tidy changes that should be on the gh-pages branch.

marcoscaceres:
@kenchris how do propose we handle this on the Chrome side? This is a breaking change. I think it’s worth doing, but we will need to convince folks (hi @mounirlamouri! 🤗)

mounirlamouri:
Is there a recommended pattern by the TAG or commonly used one from other specifications?

kenchris:
Let me talk with @slightlyoff

kenchris:
Please have a look at https://www.w3.org/2001/tag/doc/promises-guide

marcoscaceres:
@mounirlamouri wrote
> Is there a recommended pattern by the TAG or commonly used one from other specifications?

I was actually thinking of filing a bug on the TAG as this particular situation is not _exactly_ covered by either the API Design Guide or the Promises Guide.  

Regardless, the "pending promise" effectively represents the screen having changed:

```JS
const pendingPromise = new Promise(r => {
  screen.orientation.addEventListener("change", ()=> r(), { once: true });
});
```

kenchris:
> I was actually thinking of filing a bug on the TAG as this particular situation is not _exactly_ covered by either the API Design Guide or the Promises Guide.

Yes, I noticed the same. Please file

marcoscaceres:
@kenchris, the closest I get in the Promises Guide is this section:
https://www.w3.org/2001/tag/doc/promises-guide#one-time-events

> The prototypical example of such an "event" is a loaded indicator: a resource such as an image, font, or even document, could provide a loaded property that is a promise that becomes fulfilled only when the resource has fully loaded (or becomes rejected if there’s an error loading the resource).

The above suggests that the **event occurs**, and, as a result the promise is resolved and rejected based on what the result is.  

> Then, authors can always queue up actions to be executed once the resource is ready by doing resource.loaded.then(onLoaded, onFailure). 

The above follows... 

> This will work even if the resource was loaded already, queueing a microtask to execute onLoaded. This is in contrast to an event model, where if the author is not subscribed at the time the event fires, that information is lost.

... and so on... 

Anyway, will file a bug :)

marcoscaceres:
Filed https://github.com/w3ctag/promises-guide/issues/57

marcoscaceres:
@mounirlamouri, the TAG has kindly updated their [API Design doc](https://github.com/kenchris/design-principles/commit/46ab149c7da5f9d1103144add3b3207a627c8e98
) providing advice around this - thanks @kenchris for making that happen! 

Given the TAG's advice, would you be ok with changing the behavior? @Johanna-hub and I can update the implement on the Gecko side, and provide WPTests.

marcoscaceres (approved):
LGTM

Johanna-hub:
We now have a test for this https://github.com/web-platform-tests/wpt/pull/15267
@kenchris and anyone else interested, would be great if you could take a look and let us know any comments?

I have also filed a bug on bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1525582

mounirlamouri:
LGTM. I don't know if and when I would update the implementation in Chromium. Though, @Johanna-hub if you want to send a patch, I would happily review :)

marcoscaceres:
@mounirlamouri, sure, we can probably do it. Being new the Blink code base, if you can quickly just point us to the right files, we can go from there.

mounirlamouri:
Actually, I realised that I had in my backlog removing all the Android Jelly Bean support from the screen orientation code. Will probably do both at the same time.

marcoscaceres:
@mounirlamouri, ok cool. We will make sure to have good web platform tests for you at least :)

marcoscaceres:
oh, before I forget, do you want us to file a bug for this specifically on the Chrome side? Or do you have a link to your refactor bug?

mounirlamouri:
I've updated the initial message with the bug number. I have a CL ready that seems to work. Should be in soon.

mounirlamouri:
The change has landed in Blink. Hopefully it will stick. Subtle timing changes like this always worry me :)

marcoscaceres:
Ok, excellent. I’ll make sure the changes happen on the Mozilla side ASAP. Then our implementations can be in sync.

marcoscaceres (approved):
Ok, let's go with this. We need to fix this up to use task queues, however.

marcoscaceres:
Working on the Gecko patch.

marcoscaceres:
Ok, landed the changes in Gecko too. Fingers crossed it doesn't get backed out.

Issue #150, opened 2019-02-07
Definition of default orientation


Johanna-hub:
Currently default orientation is defined with the following:  

```
        A <a>document</a>'s <dfn>default orientation</dfn> is the set of
          orientations to which the screen orientation is locked when it is not
          explicitly locked by this API or any other means.
```
It would be great to get some clarification on why default orientation is a set?  From looking at the android docs it looks like the system decides on one orientation if none is specified.

@mounirlamouri Could you please explain this further?

Relates to https://github.com/w3c/screen-orientation/issues/104

mounirlamouri:
The default orientation is indeed system-specific. For example, on most Android phones, the default orientation is landscape + portrait-primary (IIRC) but it can be changed to be portrait-primary only. We could expose this information in the API but at the time we didn't want to. FWIW, Chrome's media controls use this information (if portrait-primary only) so it may make sense to expose this to the Web so websites can also implement logic that are based on the default orientation.

Johanna-hub:
I think this could make sense because at the moment the definition is a bit vague, but exposing the information would make it very useful.

@marcoscaceres What do you think?

marcoscaceres:
If we can get the defaults from the OS, and assuming they don't change, it could be good to expose them.  However, exposing these does add to device fingerprinting ... as the array's values will be device specific (though not particularly varied from one device to another). 

@mounirlamouri, how do you use the defaults in the media controls? (looking for the use case, basically).

marcoscaceres:
Rethinking this, the "set" should go away, as unlocking just leaves it to the UA/OS to handle the orientations (whatever those may be).

marcoscaceres:
The orientations really represent the orientations to which a user can then change by rotating the device. So, landscape really means that the user can change the orientation to either landscape primary and landscape secondary, but not any portrait orientation. 


I'm going to redefine things in those terms.

marcoscaceres:
Respecified in latest draft.

PR #171, opened 2019-03-11
Edit to 5.3.1. Locking to the default orientation 

Johanna-hub:
Added a bit more clarification based on #150

***
<a href="https://pr-preview.s3.amazonaws.com/Johanna-hub/screen-orientation/pull/171.html" title="Last updated on Mar 12, 2019, 5:59 PM UTC (bac9c0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/171/5cb4cfd...Johanna-hub:bac9c0d.html" title="Last updated on Mar 12, 2019, 5:59 PM UTC (bac9c0d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -722,8 +722,12 @@
             From the perspective of a <a>document</a>, locking to the
             <a>default orientation</a> is equivalent to unlocking because it
             means that it no longer has a lock applied. However, this does not
-            mean that the <a>[[\defaultOrientation]]</a> only contains the item
-            <a>any</a>.
+            mean that the <a>[[\defaultOrientation]]</a> should only contain
+            the item <a>any</a>. The <a>default orientation</a> is
+            system-specific and <a>[[\defaultOrientation]]</a> could for
+            example contain <a>portrait-primary</a> and/or
+            <a>landscape-primary</a> or the user could set to a specific
+            preference.
           </p>
         </section>
       </section>
--- a/index.html
+++ b/index.html
@@ -722,12 +722,13 @@
             From the perspective of a <a>document</a>, locking to the
             <a>default orientation</a> is equivalent to unlocking because it
             means that it no longer has a lock applied. However, this does not
-            mean that the <a>[[\defaultOrientation]]</a> should only contain
-            the item <a>any</a>. The <a>default orientation</a> is
-            system-specific and <a>[[\defaultOrientation]]</a> could for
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -722,8 +722,12 @@ <h2>
             From the perspective of a <a>document</a>, locking to the
             <a>default orientation</a> is equivalent to unlocking because it
             means that it no longer has a lock applied. However, this does not
-            mean that the <a>[[\defaultOrientation]]</a> only contains the item
-            <a>any</a>.
+            mean that the <a>[[\defaultOrientation]]</a> should only contain
```
marcoscaceres: Careful with rfc2119 keyword use here. 
```suggestion
            mean that the <a>[[\defaultOrientation]]</a> will only contain
```

index.html
```diff
@@ -722,8 +722,12 @@ <h2>
...
+            the item <a>any</a>. The <a>default orientation</a> is
```
marcoscaceres: Can I suggest we change this section to:

```HTML
        <p>
            From the perspective of a <a>document</a>, locking to the
            <a>default orientation</a> is equivalent to unlocking because it
            means that it no longer has a lock applied. 
          </p>
          <p class="note" title="Who controls the default orientation?">
            This does not
            mean that the <a>[[\defaultOrientation]]</a> will only contain
            the item <a>any</a>. The <a>default orientation</a> is
            likely device-specific and <a>[[\defaultOrientation]]</a> could, for
            example, contain <a>portrait-primary</a> and/or
            <a>landscape-primary</a>. Alternatively, the user could restrict the default orientaiton to a specific
            orientation via some OS or browser level preference for accessiblity reasons.
          </p>
```

marcoscaceres (changes requested):
Some suggestions of how we could make things more clear.

index.html
marcoscaceres: I somewhat feel that this section remain too conversational, so I'm still inclined to make it into a note (as suggested above).  Also, we should probably add a link to the web manifest section here, as it's pertinent to setting the default orientation.

Johanna-hub: Sorry, didn't see the note suggestion!  That makes sense, will do that and add in the web manifest section link

marcoscaceres (approved):
👍

PR #183, opened 2019-07-23
BREAKING CHANGE: unlock() returns a pormise

### LINKED ISSUE #104: Promise for unlock()?
I understand that this might not be possible from the note -

> unlock() does not return a Promise because it is equivalent to locking to the default orientation which might or might not be known by the user agent. Hence, the user agent can not predict what the new orientation is going to be and even if it is going to change at all.

I am working on building an app that uses WebVR. Entering VR mode requires going fullscreen and then locking orientation to landscape. This works perfectly.

However exiting this mode when phone is in portrait orientation, on tested Android devices, results in the browser bar and menu disappearing and a large white space appearing under the app. Dragging down on the screen reveals the browser bar and covers the white space, you can not then drag back up.

orientation unlock issue

I've overcome this issue using the following code
```
screen.orientation.unlock();
setTimeout( () => {
    fullscreen.exit();
}, 0 );
```
A cleaner solution would be something like -
```
screen.orientation.unlock().then( () => {
    fullscreen.exit();
});
```

marcoscaceres:
Closes #104 

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [x] Modified Web platform tests - Gecko implementation will update the tests

Implementation commitment:

 * [x] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1567805)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/183.html" title="Last updated on Jul 23, 2019, 5:52 AM UTC (9282aad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/183/da743fe...9282aad.html" title="Last updated on Jul 23, 2019, 5:52 AM UTC (9282aad)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -142,12 +142,17 @@
               show();
               updateDetails(document.getElementById("button"));
             });
+
+            async function doUnlock() {
+              await screen.orientation.unlock();
+              await document.exitFullscreen();
+            }
             &lt;/script&gt;
 
             &lt;button onclick="rotate(this)" id="button"&gt;
               Lock
             &lt;/button&gt;
-            &lt;button onclick="screen.orientation.unlock()"&gt;
+            &lt;button onclick="doUnlock()"&gt;
               Unlock
             &lt;/button&gt;
           </pre>
@@ -256,7 +261,7 @@
           [Exposed=Window]
           interface ScreenOrientation : EventTarget {
             Promise&lt;void&gt; lock(OrientationLockType orientation);
-            void unlock();
+            Promise&lt;void&gt; unlock();
             readonly attribute OrientationType type;
             readonly attribute unsigned short angle;
             attribute EventHandler onchange;
@@ -324,17 +329,9 @@
         </h2>
         <p>
           When the {{unlock()}} method is invoked, the <a>user agent</a> MUST
-          run the steps to <a>lock the orientation</a> of the <a>responsible
-          document</a> to its <a>default orientation</a>.
+          run the steps to <a>apply an orientation lock</a> of the
+          <a>responsible document</a> to its <a>default orientation</a>.
         </p>
-        <p class='note' title="Why does unlock() not return a promise?">
-          {{unlock()}} does not return a promise because it is equivalent to
-          locking to the <a>default orientation</a> which might or might not be
-          known by the <a>user agent</a>. Hence, the <a>user agent</a> can not
-          predict what the new orientation is going to be and even if it is
-          going to change at all.
-        </p>
-        <div class="issue" data-number="104"></div>
       </section>
       <section>
         <h2>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Ok, this is not exactly right... [[defaultOrientation]] is a list, but it should really just be a single value... even if the OS supports multiple.

marcoscaceres:
@mounirlamouri, could you check in Blink what you actually use for `defaultOrientation` when unlocking? From a quick search in the Blink code, it seems you also use a single value - but I wasn't able to find where it is defined (does Google have something like searchfox?). In Gecko, we use pass `eScreenOrientation_None` to the HAL. 

We could basically do the same thing in the spec by adding "default" or "none" to the `OrientationType` enum. The API itself would never return the value, but then we could pass it internally to `unlock()` to represent the "default orientations".

mounirlamouri:
The thing is for all orientation locks, the UA can figure out whether we are already in an orientation that matches the lock or whether we should wait for a change. With `unlock()` or `default`, the UA would need to know what is the system lock (in most cases, when the manifest has an orientation, we should know). If we can implement this reliably on Android and iOS, I think it would be a fine addition to the spec but I'm not sure if we want to just return immediately.

marcoscaceres:
> If we can implement this reliably on Android and iOS, I think it would be a fine addition to the spec but I'm not sure if we want to just return immediately.

Problem is that we get into situations outlined in the original bug, where screen orientation becomes racy with `exitFullScreen()` and potentially other things (see @kenchris's issue also). Even if the UA can't determine if the OS changed/unlocked orientation, it should be sufficient to know that everything possible was done by the UA to change the orientation: the task was queued, request was sent to the OS, etc.

mounirlamouri:
I would be okay to have `unlock()` return a promise but I think we should have the spec be clearer about the situation. The change as it is assumes that you can `apply an orientation lock` to the `default orientotion` which isn't possible actually. #150 kind of touch this issue. Unless we want to solve it first (maybe we do?), one way to resolve this is to simply make the `unlock()` steps async and have it return a promise when it's done so it would be clear that we are not attempting to notify that the unlock orientation change happened but that the call was registered by the system. A note in the spec making this clear would be good too.

weyyserabut:
> Closes #104
> 
> The following tasks have been completed:
> 
> * [x]  Confirmed there are no ReSpec/BikeShed errors or warnings.
> * [x]  Modified Web platform tests - Gecko implementation will update the tests
> 
> Implementation commitment:
> 
> * [ ]  WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
> * [ ]  Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
> * [x]  [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1567805)
> 
> [Preview](https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/183.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/183/da743fe...9282aad.html)

close please

weyyserabut:
close plase

PR #215, opened 2022-10-12
Rewrite privacy considerations section

### LINKED ISSUE #181: Privacy notes are overly dismissive
In Gecko, we "resist fingerprinting" when:

 * getting the type, 
 * getting the angle. 
 * orientation changes. 

We should make it clear in the privacy section that user agents are free to lie, and just return default values (just 0 for angle  and "landscape-primary"  for type).

### LINKED ISSUE #188: Should the -secondary orientations be exposed?
Are there any use cases for locking to the upside-down (portrait-secondary, landscape-secondary) orientations? Generally people would only care about landscape and portrait, and would want the device to choose the closest one to rotate to.

marcoscaceres:
Closes #188
Closes #181

Cc @smfr 

This matches [what Gecko does](https://github.com/mozilla/gecko-dev/blob/7f729f601c0b738f870ae0ed49098f9268e250f9/dom/base/ScreenOrientation.cpp#L449-L461). 

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=246511)
 * [ ] [Gecko](https://github.com/mozilla/gecko-dev/blob/7f729f601c0b738f870ae0ed49098f9268e250f9/dom/base/ScreenOrientation.cpp#L449-L461)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/215.html" title="Last updated on Oct 25, 2022, 2:31 AM UTC (afd3b18)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/215/b69e827...afd3b18.html" title="Last updated on Oct 25, 2022, 2:31 AM UTC (afd3b18)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -943,26 +943,31 @@
       </h2>
       <p>
         A screen's [=current orientation type|type=] and [=current orientation
-        angle|angle=] are a potential fingerprinting vector. To resist
-        fingerprinting (e.g., in private browsing), user agents MAY:
+        angle|angle=] are a potential fingerprinting vector. The following
+        mitigation help protect a user's privacy by not revealing how a device
+        is being held, and also prevents the [=secondary=] orientation type and
+        associated angles from being user for fingerprinting purposes.
       </p>
-      <ol>
-        <li>Restrict the value return by the {{ScreenOrientation/type}}
-        attribute to {{OrientationType/"portrait-primary"}} or
-        {{OrientationType/"landscape-primary"}} to match the screen's aspect
-        ratio.
+      <p>
+        To resist fingerprinting (e.g., in private browsing), user agents MAY:
+      </p>
+      <ul>
+        <li>Depending on the screen's aspect ratio, only return either
+        {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-primary"}} for the value of the
+        {{ScreenOrientation/type}} attribute.
         </li>
-        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
-        attribute.
+        <li>For any [=natural=] orientation ([=portrait=] or [=landscape=]),
+        always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute. And for any rotated/opposite orientation, always return `90`
+        for the value of the {{ScreenOrientation/angle}} attribute.
         </li>
         <li>If the screen orientation changes, only fire the
           <a data-link-for="ScreenOrientation">change</a> event when the
           [=current orientation type=] changes from [=portrait=] to
-          [=landscape=], or vice versa. This both protects user's privacy by
-          not revealing how a device is being held, and prevents using the
-          [=secondary=] orientation as a fingerprinting vector.
+          [=landscape=], or vice versa.
         </li>
-      </ol>
+      </ul>
     </section>
     <section id="conformance"></section>
     <section id="idl-index" class="appendix"></section>
--- a/index.html
+++ b/index.html
@@ -949,14 +949,14 @@
       <ol>
         <li>Restrict the value return by the {{ScreenOrientation/type}}
         attribute to {{OrientationType/"portrait-primary"}} or
-        {{OrientationType/"landscape-secondary"}} so to match the screen's
+        {{OrientationType/"landscape-secondary"}} to match the screen's
         aspect ratio.
         </li>
         <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
         attribute.
         </li>
-        <li>If the screen orientation changes, not fire the
-        {{ScreenOrientation/onchange}} event to reveal a change to a
+        <li>If the screen orientation changes, not fire the <a data-link-for=
+        "ScreenOrientation">change</a> event to reveal a change to a
         [=secondary=] orientation.
         </li>
       </ol>
--- a/index.html
+++ b/index.html
@@ -967,10 +967,11 @@
         To resist fingerprinting (e.g., in private browsing), user agents MAY:
       </p>
       <ul>
-        <li>Depending on the screen's aspect ratio, only return either
+        <li>Restrict the possible return values of the
+        {{ScreenOrientation/type}} getter to
         {{OrientationType/"portrait-primary"}} or
-        {{OrientationType/"landscape-primary"}} for the value of the
-        {{ScreenOrientation/type}} attribute.
+        {{OrientationType/"landscape-primary"}}. The screen aspect ratio
+        determines which is returned.
         </li>
         <li>For any [=natural=] orientation ([=portrait=] or [=landscape=]),
         always return `0` for the value of the {{ScreenOrientation/angle}}
--- a/index.html
+++ b/index.html
@@ -937,9 +937,9 @@
         {{OrientationType/"landscape-primary"}}. The screen aspect ratio
         determines which is returned.
         </li>
-        <li>When the [=Screen/current orientation type=] matches
-        {{ScreenOrientation/[[initialType]]}}, the {{ScreenOrientation/angle}}
-        getter MUST return `0`. Return `90` otherwise.
+        <li>If the [=Screen/current orientation type=] matches
+        {{ScreenOrientation/[[initialType]]}}, return `0` for the the
+        {{ScreenOrientation/angle}} getter. Otherwise, return `90`.
         </li>
         <li>If the screen orientation changes, only fire the
           <a data-link-for="ScreenOrientation">change</a> event when the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -937,20 +937,29 @@ <h2>
...
-          fingerprinting vector.
-        </p>
-      </section>
+      <p>
+        A screen's [=current orientation type|type=] and [=current orientation
+        angle|angle=] is a potential fingerprinting vector. To resist
+        fingerprinting (e.g., in private browsing), user agents MAY:
+      </p>
+      <ol>
+        <li>restrict the value return by the {{ScreenOrientation/type}}
```
```suggestion
        <li>Restrict the value return by the {{ScreenOrientation/type}}
```

index.html
```diff
@@ -937,20 +937,29 @@ <h2>
...
+        fingerprinting (e.g., in private browsing), user agents MAY:
+      </p>
+      <ol>
+        <li>Restrict the value return by the {{ScreenOrientation/type}}
+        attribute to {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-secondary"}} so to match the screen's
+        aspect ratio.
+        </li>
+        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute
```
```suggestion
        attribute.
```

index.html
```diff
@@ -937,20 +937,29 @@ <h2>
...
-        </h3>
-        <p>
-          The screen orientation type and angle of the device can be accessed
-          with the API specified in this document, and can be a potential
-          fingerprinting vector.
-        </p>
-      </section>
+      <p>
+        A screen's [=current orientation type|type=] and [=current orientation
+        angle|angle=] is a potential fingerprinting vector. To resist
```
annevk: Nit: are?

index.html
```diff
@@ -937,20 +937,29 @@ <h2>
...
+        attribute to {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-secondary"}} to match the screen's
+        aspect ratio.
+        </li>
+        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute.
+        </li>
+        <li>If the screen orientation changes, not fire the <a data-link-for=
+        "ScreenOrientation">change</a> event to reveal a change to a
+        [=secondary=] orientation.
```
annevk: Wouldn't this encourage battery-expensive polling instead? If you support rotating in a way that changes width/height of the viewport I think we should require that something changes. But limiting it two values for type and angle each seems reasonable.

index.html
```suggestion
        angle|angle=] are a potential fingerprinting vector. To resist
```

index.html
marcoscaceres: > Wouldn't this encourage battery-expensive polling instead?

Sorry, I might need to word this better. The answer is "no", because the UA wouldn't fire events when switching from "X-primary" to "X-secondary". Only if "X" changes, would the event fire, but it would always report as "X-primary". 

The reason the orientation change event does fire, is that the screen width/height would change, which is already observable either polling `screen`'s attribute or simply by `matchMedia("(orientation: landscape)")`.

marcoscaceres:
Thanks for the feedback @annevk. I updated the event recommendation. Is that more clear?

index.html
```diff
@@ -937,20 +937,32 @@ <h2>
...
+        </li>
+        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute.
+        </li>
+        <li>If the screen orientation changes, only fire the
+          <a data-link-for="ScreenOrientation">change</a> event when the
+          [=current orientation type=] changes from [=portrait=] to
+          [=landscape=], or vice versa. This both protects user's privacy by
+          not revealing how a device is being held, and prevents using the
+          [=secondary=] orientation as a fingerprinting vector.
```
annevk: Should this say "other" since above it wants to use "landscape-secondary"? Or is that a typo above?

index.html
```diff
@@ -937,20 +937,32 @@ <h2>
...
+        A screen's [=current orientation type|type=] and [=current orientation
+        angle|angle=] are a potential fingerprinting vector. To resist
+        fingerprinting (e.g., in private browsing), user agents MAY:
+      </p>
+      <ol>
+        <li>Restrict the value return by the {{ScreenOrientation/type}}
+        attribute to {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-secondary"}} to match the screen's aspect
+        ratio.
+        </li>
```
annevk, line 946: This still reads a little weird. Can it return one of those or both, but what it returns depends on the aspect ratio? (Presumably the latter.)

index.html
```diff
@@ -937,20 +937,32 @@ <h2>
...
+        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute.
+        </li>
```
annevk, line 950: Shouldn't this be 0 or 90 to match the thing above? Or maybe angle is more complicated in which case this might warrant some further documentation here and in the angle domintro as this would mean you cannot rely on it changing.

annevk (commented):
Thanks, that helps! Only nits remain.

index.html
marcoscaceres: oh lol, yes, that was a typo above.

index.html
```diff
@@ -937,20 +937,32 @@ <h2>
...
-      </section>
+      <p>
+        A screen's [=current orientation type|type=] and [=current orientation
+        angle|angle=] are a potential fingerprinting vector. To resist
+        fingerprinting (e.g., in private browsing), user agents MAY:
+      </p>
+      <ol>
+        <li>Restrict the value return by the {{ScreenOrientation/type}}
+        attribute to {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-secondary"}} to match the screen's aspect
```
```suggestion
        {{OrientationType/"landscape-primary"}} to match the screen's aspect
```

index.html
marcoscaceres: Yeah, I was chewing on this also... I like the idea of returning `90` for landscape-primary if it's obvious (potentially derived by UA string) that the device has a natural orientation of portrait. Otherwise, it could be used to detect if the user is in private browsing mode.

marcoscaceres:
Hopefully this is a bit clearer now...

index.html
annevk: Maybe something along these lines:

> Restrict the possible return values of the `type` getter to portrait and landscape. The screen aspect ratio determines which is returned.

index.html
```diff
@@ -957,20 +957,37 @@ <h2>
...
+      <ul>
+        <li>Depending on the screen's aspect ratio, only return either
+        {{OrientationType/"portrait-primary"}} or
+        {{OrientationType/"landscape-primary"}} for the value of the
+        {{ScreenOrientation/type}} attribute.
+        </li>
+        <li>For any [=natural=] orientation ([=portrait=] or [=landscape=]),
+        always return `0` for the value of the {{ScreenOrientation/angle}}
+        attribute. And for any rotated/opposite orientation, always return `90`
+        for the value of the {{ScreenOrientation/angle}} attribute.
```
annevk: Okay, so this means we will be exposing the natural direction of the device. Or is the natural direction simply the direction the user used when navigating to the site?

index.html
marcoscaceres: used your suggestion.

index.html
marcoscaceres: Oh! I quite like the idea of taking the orientation to which the page was navigated and using it to lie (i.e., and not expose the natural orientation). 

No browser does the above today, which is why I spec'ed it based on "natural" orientation, but that doesn't mean they couldn't.

index.html
```diff
@@ -898,30 +911,40 @@ <h2>
...
+        {{ScreenOrientation/type}} getter to
         {{OrientationType/"portrait-primary"}} or
-        {{OrientationType/"landscape-primary"}}.
+        {{OrientationType/"landscape-primary"}}. The screen aspect ratio
+        determines which is returned.
         </li>
-        <li>Always return `0` for the value of the {{ScreenOrientation/angle}}
-        attribute.
+        <li>If the [=Screen/current orientation type=] matches
+        {{ScreenOrientation/[[initialType]]}}, return `0` for the the
```
```suggestion
        {{ScreenOrientation/[[initialType]]}}, return `0` for the
```

marcoscaceres:
Merging as feedback has been addressed. Happy to refine this a bit more as a followup.

PR #231, opened 2022-10-27
Allow unlock() to return a promise

### LINKED ISSUE #104: Promise for unlock()?

marcoscaceres:
Closes #104 

 * [ ] Modified Web platform tests (link to pull request)

Implementation commitment:

 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/231.html" title="Last updated on Oct 21, 2025, 4:14 AM UTC (dda88e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/231/3e5af26...dda88e8.html" title="Last updated on Oct 21, 2025, 4:14 AM UTC (dda88e8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -515,7 +515,7 @@
           <li>If |document| has the [=sandboxed orientation lock browsing
           context flag=] set, return [=a promise resolved with=] `undefined`.
           </li>
-          <li>If screen's [=Screen/active orientation lock=] is `null`, return
+          <li>If screen's [=Screen/active orientation lock=] is `null`,
           return [=a promise resolved with=] `undefined`.
           </li>
           <li data-tests="lock-basic.html">If |document|'s
--- a/index.html
+++ b/index.html
@@ -91,6 +91,11 @@
             updateLockButton(lockButton);
           }
 
+          async function unlock() {
+            await screen.orientation.unlock();
+            updateLockButton();
+          }
+
           screen.orientation.addEventListener("change", updateLockButton);
 
           window.addEventListener("load", updateLockButton);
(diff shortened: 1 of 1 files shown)


annevk:
Doesn't this have the problem that the promise is resolved before the lock might have been applied? (As that happens in parallel.)

marcoscaceres:
It shouldn't (though maybe I didn't word it right?). That definitely should never happen: in parallel, (un)locking happens, maybe the event fires if the orientation changed, then the promise is resolved.

annevk:
I think I understand, I got confused due to it sometimes early returning a successful promise.

marcoscaceres:
Instead of this, rely on exit fullscreen to automatically unlock.

PR #232, opened 2022-10-27
Prevent hidden documents from locking orientation

### LINKED ISSUE #221: Should hidden documents be allowed to change the orientation?
I'm thinking we should disallow hidden documents from changing the orientation. 

I can't think of a use case whereby this should be allowed.

Together with #210 (require a gesture) that should vastly simplify when/how this API can be used.

marcoscaceres:
Closes #221 

 * [x] [Modified Web platform tests](https://github.com/web-platform-tests/wpt/pull/36734)

Implementation commitment:

 * [x] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=247248) - [patch](https://github.com/WebKit/WebKit/pull/6283)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=1802823)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/232.html" title="Last updated on Feb 15, 2023, 5:35 AM UTC (638c545)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/232/9dc8f9d...638c545.html" title="Last updated on Feb 15, 2023, 5:35 AM UTC (638c545)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -470,13 +470,13 @@
           risk, return [=a promise rejected with=] a {{"SecurityError"}}
           {{DOMException}} and abort these steps.
           </li>
+          <li>If |document|'s [=Document/visibility state=] is "hidden", return
+          [=a promise rejected with=] an {{"SecurityError"}} {{DOMException}}
+          and abort these steps.
+          </li>
           <li>If the [=user agent=] does not support locking the screen
           orientation to |orientation|, return [=a promise rejected with=] a
           {{"NotSupportedError"}} {{DOMException}} and abort these steps.
-          </li>
-          <li>If |document|'s [=Document/visibility state=] is "hidden", return
-          [=a promise rejected with=] an {{"NotAllowedError"}} {{DOMException}}
-          and abort these steps.
           </li>
           <li data-tests="lock-basic.html">If |document|'s
           {{Document/[[orientationPendingPromise]]}} is not `null`, [=reject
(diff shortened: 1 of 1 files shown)


annevk:
I guess this doesn't apply to `unlock()` because that shouldn't be possible? Maybe assert it? (Although maybe it's possible with manifest apps?)

marcoscaceres:
> I guess this doesn't apply to unlock() because that shouldn't be possible? Maybe assert it? (Although maybe it's possible with manifest apps?)

I was thinking it shouldn't be possible because to hide the page/app it once locked, a user would need to exit fullscreen. But yeah, the assumption doesn't hold when there are multiple virtual desktops. 

It may be safer just to prevent it in both cases.

annevk (approved):
Probably makes sense to split out the safety checks into a shared algorithm at some point.

marcoscaceres:
Prepared test https://github.com/web-platform-tests/wpt/pull/36734  

[Filed WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=247248). 

@saschanaz or @makotokato, @michaelwasserman, could you please file bugs in your respective engine repos if you support the change?

marcoscaceres:
Sent a patch for WebKit https://github.com/WebKit/WebKit/pull/6283

marcoscaceres:
This change has been [merged in WebKit](https://github.com/WebKit/WebKit/pull/6283).

makotokato:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1802823

makotokato:
BTW, https://github.com/web-platform-tests/wpt/pull/36734's test (`hidden_document.html`) is still unclear for me.  "hidden documents must not unlock the screen orientation" and "Once maximized, a minimized window can lock or unlock the screen orientation again" don't set pre-lock condition (All browser requires full screen). So I think the latest version of these tests are always failure due to no fullscreen.

Does it forget adding `documenttElement.requestFullscreen` before `orientation.lock` call?

makotokato:
And, why don't we care hidden state in "8.3 Applying an orientation lock" section like fully active if it is in parallel case?

marcoscaceres:
@makotokato, apologies, just coming back to this. 

Thanks for pointing out the issues with the tests and in section "8.3 Applying an orientation lock". I think that was just an oversight. Having a look now.

marcoscaceres:
@makotokato, about the tests, I've gone over them again and you are right... they were making incorrect assumptions. For example, on macOs at least, it appears one cannot minimize a window once it's in fullscreen. One needs to first exit fullscreen and then the window can be minimized. 

Anyway, I'll delete those two tests. 

> And, why don't we care hidden state in "8.3 Applying an orientation lock" section like fully active if it is in parallel case?

Yeah, good catch. I think instead what needs to happen there is that if the preconditions are not longer met (which is remaining visible and in fullscreen).

marcoscaceres:
@makotokato, I've filed https://github.com/w3c/screen-orientation/issues/243 to follow up on your comment above. I'll draft up a separate pull request for that. 

About the tests, I sent https://github.com/web-platform-tests/wpt/pull/38511

PR #236, opened 2022-10-28
Editorial: rewrite parallel steps

### LINKED ISSUE #234: "reject and nullify the current lock promise" and tasks
I don't really understand why this queues tasks, until I saw that it also had an "in parallel" caller. However, that caller should probably queue a task to run this as this algorithm inspects document state that's not going to be available in parallel, right?

marcoscaceres:
Closes #234

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/236.html" title="Last updated on Nov 1, 2022, 7:47 PM UTC (37c686e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/236/ebe68e0...37c686e.html" title="Last updated on Nov 1, 2022, 7:47 PM UTC (37c686e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -723,12 +723,13 @@
             </li>
           </ol>
         </li>
-        <li>Run the following sub-steps [=in parallel=]. If |document| stops
-        being [=Document/fully active=] while [=in parallel=], and if the
-        |document|'s {{Document/[[orientationPendingPromise]]}} is not `null`,
-        [=reject and nullify the current lock promise=] of |document| with an
-        {{"AbortError"}}:
+        <li>Run the following sub-steps [=in parallel=]:
           <ol>
+            <li>If |document| has stopped being [=Document/fully active=], and
+            if the |document|'s {{Document/[[orientationPendingPromise]]}} is
+            not `null`, [=reject and nullify the current lock promise=] of
+            |document| with an {{"AbortError"}}.
+            </li>
             <li>If |orientation| is `null`, [=unlock the screen orientation=].
             </li>
             <li>Otherwise, attempt to [=lock the screen orientation=] to
--- a/index.html
+++ b/index.html
@@ -701,13 +701,9 @@
       <p>
         When steps require to <dfn>apply orientation lock</dfn> of
         {{OrientationLockType?}} |orientation| to {{Document}} |document|, the
-        [=user agent=] MUST perform do the following steps [=in parallel=]:
+        [=user agent=] MUST perform do the following steps:
       </p>
-      <ol>
-        <li>If |document| stops being [=Document/fully active=] while [=in
-        parallel=], [=reject and nullify the current lock promise=] of
-        |document| with an {{"AbortError"}}.
-        </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -727,52 +723,69 @@ <h2>
...
-        <li>Otherwise, attempt to [=lock the screen orientation=] to
-        |orientation|. Depending on platform conventions, change how the
-        viewport is drawn to match |orientation|. If the attempt fails due to
-        previously-established user preference, or platform limitation, or any
-        other reason:
+        <li>Run the following sub-steps [=in parallel=]. If |document| stops
+        being [=Document/fully active=] while [=in parallel=], and if the
+        |document|'s {{Document/[[orientationPendingPromise]]}} is not `null`,
+        [=reject and nullify the current lock promise=] of |document| with an
+        {{"AbortError"}}:
```
annevk: Technically speaking this is not something you can do. You'd have to periodically queue tasks to make this check. But realistically this should all happen so fast that you should just check this once the in parallel algorithm has done its in parallel thing and queues a task to continue?

marcoscaceres: Hmm, interesting... It might be a Gecko-ism, but there it was possible to get notifications of the document changing state (e.g., in payment request,  it was used this to [invalidate various promises](https://searchfox.org/mozilla-central/source/dom/payments/PaymentRequest.cpp#1209)). 

I'll move this though, as you suggested, as you are right that everything happens in quick succession (unlike payment request, where the payment sheet is indefinitely waiting for user input).

marcoscaceres: Thinking out loud...  this is the only case we can realistically capture:

```JS
const iframe = document.querySelector("iframe");
const p = iframe.contentWindows.screen.orientation.lock("landscape");
iframe.remove();
// p aborts
```

So, the check should happen as the very first thing after going "in parallel".

annevk: It's probably possible to design a system whereby you get notified, but that would then be some kind of thing you register for in the HTML standard and not something you can access when in parallel.

marcoscaceres: Yeah, understood. Do you think the intention is clear enough though? It's not too dissimilar to, for instance, in img [`decode()`](https://html.spec.whatwg.org/#dom-img-decode) method. It also has similar wording that happens in parallel:

> Otherwise, [in parallel](https://html.spec.whatwg.org/#in-parallel), wait for one of the following cases to occur, and perform the corresponding actions: 
> This [img](https://html.spec.whatwg.org/#the-img-element) element's [node document](https://dom.spec.whatwg.org/#concept-node-document) stops being [fully active](https://html.spec.whatwg.org/#fully-active)
>... other cases...
>Reject promise with an ["EncodingError"](https://webidl.spec.whatwg.org/#encodingerror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

annevk: I guess it's good enough, but you should at least queue a task. That's an obvious bug in `decode()` we don't have to mimic.

PR #246, opened 2023-04-02
Fix error in example usage

### LINKED ISSUE #245: Example of usage is incorrect
Could there be an error in Example 1? `getOppositeOrientation()` uses the line `screen.orientation.startsWith("portrait")` as if screen.orientation returns a string, when it actually returns an object with the type (so the command worked for me if i wrote `screen.orientation.type.startsWith("portrait")`. I would appreciate guidance on this matter if I'm wrong.

NotWoods:
Closes #245

@Maferep caught that the example treats `screen.orientation` as a string instead of an object. Updates the example to use `.type` instead.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/246.html" title="Last updated on Apr 2, 2023, 8:24 PM UTC (1c9e0e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/246/c0fcb39...1c9e0e6.html" title="Last updated on Apr 2, 2023, 8:24 PM UTC (1c9e0e6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -80,6 +80,7 @@
           function getOppositeOrientation() {
             return screen
               .orientation
+              .type
               .startsWith("portrait") ? "landscape" : "portrait";
           }
 


marcoscaceres (approved):
Oops! thanks!!

PR #250, opened 2023-04-13
Editorial: fix queue a task to fire change an event

### LINKED ISSUE #244: Screen orientation change need to queue a task to fire an event properly
The screen orientation change steps need to be rewritten to properly queue a task to fire the "change" event, which updates the type/angle of each document's `ScreenOrientation` as part of a task. 

As currently spec'ed, it's not quite correct (though the intent is). 

This would be an editorial change. 

Note to self: 

- [x] Update screen orientation change steps
- [x] Update Handling page visibility changes section.

marcoscaceres:
Closes #244

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/250.html" title="Last updated on May 1, 2023, 5:21 AM UTC (fedbc6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/250/b476e40...fedbc6a.html" title="Last updated on May 1, 2023, 5:21 AM UTC (fedbc6a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -834,9 +834,9 @@
         |screenOrientation|'s {{ScreenOrientation/[[angle]]}}, abort these
         steps.
         </li>
-        <li>[=Queue a task=] on the [=user interaction task source=] with
-        |document|'s [=relevant global object=] to perform the following
-        substeps:
+        <li>[=Queue a global task=] on the [=user interaction task source=]
+        with |document|'s [=relevant global object=] to perform the following
+        steps:
           <ol>
             <li>Set |screenOrientation|'s {{ScreenOrientation/[[angle]]}} to
             |angle|.
@@ -853,10 +853,8 @@
         |document|'s [=Document/descendant navigables=]'s [=navigable/active
         documents=], if any, in tree order.
         </li>
-        <li>[=Set/For each=] |doc:Document| in |descendantDocs|, [=queue a
-        task=] on the [=user interaction task source=] with |doc|'s [=relevant
-        global object=] to run the [=screen orientation change steps=] with
-        |doc|.
+        <li>[=Set/For each=] |doc:Document| in |descendantDocs|, run the
+        [=screen orientation change steps=] with |doc|.
         </li>
       </ol>
       <h2>
--- a/index.html
+++ b/index.html
@@ -834,22 +834,29 @@
         |screenOrientation|'s {{ScreenOrientation/[[angle]]}}, abort these
         steps.
         </li>
-        <li>Set |screenOrientation|'s {{ScreenOrientation/[[angle]]}} to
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -834,22 +834,29 @@ <h2>
...
         steps.
         </li>
-        <li>Set |screenOrientation|'s {{ScreenOrientation/[[angle]]}} to
-        |angle|.
-        </li>
-        <li>Set |screenOrientation|'s {{ScreenOrientation/[[type]]}} to |type|.
-        </li>
-        <li>[=Fire an event=] named "<a data-link-for=
-        "ScreenOrientation">change</a>" at |screenOrientation|.
+        <li>[=Queue a task=] on the [=user interaction task source=] with
```
annevk: Queue a task is the legacy algorithm. You want one of the ones that takes more arguments, such as queue a global task.

index.html
```diff
@@ -834,22 +834,29 @@ <h2>
...
+        |document|'s [=relevant global object=] to perform the following
+        substeps:
```
annevk: We typically say "steps" these days.

index.html
```diff
@@ -834,22 +834,29 @@ <h2>
...
         |document|'s [=Document/descendant navigables=]'s [=navigable/active
         documents=], if any, in tree order.
         </li>
         <li>[=Set/For each=] |doc:Document| in |descendantDocs|, [=queue a
-        global task=] on the [=DOM manipulation task source=] with |doc|'s
-        [=relevant global object=] to run the [=screen orientation change
-        steps=] with |doc|.
+        task=] on the [=user interaction task source=] with |doc|'s [=relevant
+        global object=] to run the [=screen orientation change steps=] with
+        |doc|.
```
annevk: This seems like it's queueing a task to queue a task. That seems bad? Perhaps "screen orientation change steps" shouldn't queue a task itself after all?

marcoscaceres: yeah, you are right... will just iterate over the docs.

annevk (approved):
1. This is not editorial, so you want to address that in the commit message when landing.
2. This matches the accuracy we have in other specifications, but it's not the full answer. In particular when multiple documents are in the same agent, implementations most likely use a single task for all of them (and if they don't they really should consider that).

annevk:
I missed that this needs rebasing. If that changes things significantly I'd be happy to take another look.

marcoscaceres:
Thanks @annevk. Dropped "Editorial" from the commit message. 

I'll do a follow up and some investigation on: "In particular when multiple documents are in the same agent, implementations most likely use a single task for all of them (and if they don't they really should consider that)."

PR #251, opened 2023-04-13
Editorial: combine safety check steps into new section

### LINKED ISSUE #237: Split out the safety checks into a shared algorithm
For .lock() and .unlock(), it would be good to have shared steps to link to for where safety checks are done.

marcoscaceres:
Closes #237

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/pull/251.html" title="Last updated on May 4, 2023, 4:57 AM UTC (9a61afc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/screen-orientation/251/2f55594...9a61afc.html" title="Last updated on May 4, 2023, 4:57 AM UTC (9a61afc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -481,8 +481,9 @@
           [=sandboxed orientation lock browsing context flag=] set, return a
           {{"SecurityError"}} {{DOMException}}.
           </li>
-          <li>If |document|'s [=Document/visibility state=] is "hidden", return
-          a {{"SecurityError"}} {{DOMException}}.
+          <li data-tests="hidden_document.html">If |document|'s
+          [=Document/visibility state=] is "hidden", return a
+          {{"SecurityError"}} {{DOMException}}.
           </li>
         </ol>
       </section>
--- a/index.html
+++ b/index.html
@@ -408,9 +408,9 @@
           <li>Let |document:Document| be [=this=]'s [=relevant global
           object=]'s [=associated `Document`=].
           </li>
-          <li>Run the [=common safety checks=] with |document|. If it returns a
-          {{DOMException}}, return [=a promise rejected with=] that
-          {{DOMException}} and abort these steps.
+          <li>Run the [=common safety checks=] with |document|. If an
+          [=exception=] is [=exception/throw|thrown=], return [=a promise
+          rejected with=] that exception and abort these steps.
           </li>
           <li>If the [=user agent=] does not support locking the screen
           orientation to |orientation|, return [=a promise rejected with=] a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -533,6 +516,29 @@ <h2>
...
         </p>
       </section>
+      <section>
+        <h2>
+          Common Safety Checks
+        </h2>
+        <p>
+          The <dfn>common safety checks</dfn> for a {{Document}}
+          |document:Document| are the following steps:
+        </p>
```
annevk, line 474: You should make these throw an exception. Web IDL will do the conversion from a thrown exception to a rejected promise. Returning an exception is very weird.

Alternatively you could return a(n) (string) enum and branch on that.

marcoscaceres: Ok, yeah, good point... will make them throw directly.

marcoscaceres: Ok, sooooo... these all throw now. And yes, the IDL layer would automatically wrap the exceptions. 

However, just from the sake of clarity and consistency, I'd still like to use "Return a promise rejected with X" for `.lock()`. I've had instances in the past of reviewers getting confused about methods throwing even though they always return a promise. 

Similarly, for .unlock(), just for clarity, I have it re-throw and abort the steps.

index.html
```diff
@@ -481,6 +464,29 @@ <h2>
...
+          {{"InvalidStateError"}} {{DOMException}}.
+          </li>
+          <li data-tests="lock-sandboxed-iframe.html">If |document| has the
+          [=sandboxed orientation lock browsing context flag=] set,
+          [=exception/throw=] {{"SecurityError"}} {{DOMException}}.
+          </li>
+          <li data-tests="hidden_document.html">If |document|'s
+          [=Document/visibility state=] is "hidden", [=exception/throw=]
+          {{"SecurityError"}} {{DOMException}}.
+          </li>
```
annevk, line 487: It strikes me that if you adopt https://github.com/whatwg/html/commit/2db564c6902ace8c2539560e508ce22c2948c4d8 here (and I think you should in a follow-up) you won't get to distinguish exceptions, but that's probably okay.

index.html
annevk: I guess that's okay, but I'd rather we explain the mechanism in a note so implementations with correct binding layers don't do this redundantly as well.

annevk (approved):

index.html
marcoscaceres: Yeah, that's ok. And yeah, will do that as a followup.

index.html
marcoscaceres: Yeah, it would be way cleaner. I'd also rather just be writing: 

* If X throw {{WhateverError}}.
* If Y throw {{OtherError}}.

And so on... It feels like WebIDL's [=exception/throw=] should include such a note. Currently, it's not super clear what should happen and every example in WebIDL is using [=return a promise rejected with=]. 

I'll raise an issue on WebIDL.

marcoscaceres: Anyone interested https://github.com/whatwg/webidl/issues/1301

